PR #2, opened 2020-06-15
CQLHS-33: added dependencies to README & package.json

jonathanchaffer:
We'll add more in the future but these are some dependencies we will likely use.


jonathanchaffer:
@charkour That can probably come later since we won't really mess with styling until we have final design comps

PR #8, opened 2020-06-16
ESLint Config

charkour:
Setup ESLint configuration with AirBnB's Javascript Style Guide and the recommended React, React-Hooks, and TypeScript Configuration. Changed rules to not conflict with Prettier config. VSCode will auto-fix fixable ESLint errors.

Updated readme.


charkour:
Merging into develop so all branches have eslintrc in the future.

PR #9, opened 2020-06-16
Added navbar and sign out confirmation modal

jonathanchaffer:
Navbar still needs styling. Confirmation modal is reusable — you can pass in text, onConfirm, onCancel, etc. as props


ClientApp/src/components/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+  text: string;
+  isShowing: boolean;
+  onConfirm: () => void;
+  onCancel: () => void;
+  confirmText?: string;
+  cancelText?: string;
+  variant?: "primary" | "danger";
+}
+
+export const ConfirmModal = (props: IConfirmModalProps) => {
```
charkour, line 15: In the future, ESLint will want you to use object deconstruction so you don't need to write `props.attribute` over and over. Instead, it would be `attribute`

PR #15, opened 2020-06-17
Feature/cqlhs 48

jonathanchaffer:
- add redirect to dashboard from login page (currently uses a placeholder async function to simulate logging in)
- add a reusable error handling component & hook


jonathanchaffer:
So we can use the ErrorModal on new stories. Also, CQLHS-54 and CQLHS-48 are subtasks of different stories so I probably should've put them on separate branches.

PR #30, opened 2020-06-19
Style Login

charkour:
- Match design comps
- Make Navigation dynamically show
- Remove "forgot password" feature


ClientApp/src/components/LoginPage/LoginPage.scss
```diff
@@ -0,0 +1,7 @@
+.LoginPage {
+  margin-top: 15vh;
+}
+
+Label {
+  font-weight: 500;
+}
```
jonathanchaffer, line 7: I would change this to `.form-label` and move to `index.scss` so it's used everywhere

ClientApp/src/components/LoginPage/LoginPage.scss
```diff
@@ -0,0 +1,7 @@
+.LoginPage {
+  margin-top: 15vh;
+}
```
jonathanchaffer, line 3: Maybe instead of a static offset from the top of the screen, the form could be centered vertically?

ClientApp/src/components/LoginPage/LoginPage.tsx
```diff
@@ -19,24 +21,32 @@ export const LoginPage: React.FC = () => {
...
+              type="password"
+            />
+          </Form.Group>
+        </Container>
         <Form.Group controlId="submit">
           {isLoading ? (
             <Spinner animation="border" variant="primary" />
           ) : (
             <Button
+              className="mt-3"
```
jonathanchaffer, line 49: I would apply `mb-4` to the password `Form.Group`. Then the margin gets applied both to the button and the spinner

ClientApp/src/components/LoginPage/LoginPage.tsx
```diff
@@ -19,24 +21,32 @@ export const LoginPage: React.FC = () => {
...
+          </Form.Group>
+          <Form.Group controlId="password">
+            <Form.Label>Password</Form.Label>
+            <Form.Control
+              disabled={isLoading}
+              placeholder="Password"
+              type="password"
+            />
+          </Form.Group>
+        </Container>
```
jonathanchaffer, line 43: Any reason why the `Container` is closed here instead of after the submit `Form.Group`?

ClientApp/src/components/LoginPage/LoginPage.tsx
```diff
@@ -19,24 +21,32 @@ export const LoginPage: React.FC = () => {
...
-        </Form.Group>
-        <Form.Group controlId="password">
-          <Form.Label>Password</Form.Label>
-          <Form.Control disabled={isLoading} placeholder="Password" type="password" />
-        </Form.Group>
-        <Form.Group controlId="forgot">
-          <a href="/login">I forgot my password</a>
+          <h1 className="display-4 mb-5">Sign in to Benchmark</h1>
         </Form.Group>
+        <Container className="w-75 p-0 m-0">
```
jonathanchaffer, line 26: This can probably be a `div` instead of a `Container`

ClientApp/src/components/LoginPage/LoginPage.scss
charkour: Good call, I was trying to match the design comps _too_ much.

charkour: It looks a little weird with vertical centering so I prefer some sort of top padding

ClientApp/src/components/LoginPage/LoginPage.tsx
charkour: Thanks, I overlooked this

ClientApp/src/components/LoginPage/LoginPage.tsx
charkour: I wrapped the text inputs in a `Container` so they could be left-aligned to the parent `div`.

PR #33, opened 2020-06-22
editable new-project page

jonathanchaffer:
- Adds editable inputs
- Adds ability to add links to a project
- Values saved to component state so they can be saved to backend later
- No styling yet
- Still need searchable employee list


ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+enum ProjectRole {
+  CLIENT = "Client",
+  CSM = "Client Success Manager",
+  DESIGNER = "Designer",
+  DEVELOPER = "Developer",
+  PROJECT_MANAGER = "Project Manager",
+  QA = "Quality Assurance",
+  STRATEGIST = "Strategist",
+  TECH_LEAD = "Tech Lead",
+}
```
charkour, line 29: This eventually might get pulled out to a DB table so an Admin can update the roles that are in the system when making projects, but it is good here for now.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+export enum LinkType {
+  CONFLUENCE = "Confluence Page",
+  GITHUB = "GitHub Project",
```
charkour, line 33: Most projects at CQL are not stored in GitHub repos, so the ability to change the name of links would be important. Maybe you could keep these as defaults, but you certainly don't want to lock them into only these choices.

ClientApp/.eslintrc.js
```diff
@@ -78,12 +78,13 @@ module.exports = {
     "react/jsx-sort-props": "error",
 
     // More typescript specific rules
-    "@typescript-eslint/no-magic-numbers": "error",
+    "@typescript-eslint/no-magic-numbers": "warn",
```
charkour, line 81: Yeah, this rule is annoying, it can be taken out if you wish.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+  JIRA = "Jira Project",
+  RETRO = "Project Retro",
+}
+
+// TODO: add calendar icon button to datepickers
+export const ProjectEditor: React.FC = () => {
+  const [startDate, setStartDate] = useState<Date | null>(null);
+  const [endDate, setEndDate] = useState<Date | null>(null);
```
charkour, line 41: `startDate` and `endState` are not used. You should add them to their corresponding `DatePicker` components using `value={moment(date).format("MM/DD/YYYY")}`. 
The default state for dates should be set to something that is not null. I recommend leaving the default empty until we get info from the DB: `useState<Date | null>();`

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+  const [teamMembers, setTeamMembers] = useState<TeamMember[]>([]);
+  const nextId = (): number => {
+    let maxId = -1;
+    teamMembers.forEach((teamMember) => {
+      if (teamMember.id > maxId) maxId = teamMember.id;
+    });
+    return maxId + 1;
+  };
```
charkour, line 50: This will probably get changed to be the unique employee ID when we get the DB working. Or we might add another field for the ID since this is acting like a key it seems.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+            <Form.Group
+              className="d-flex flex-column"
+              controlId="review-deadline"
+            >
+              <Form.Label>
+                <h5>Review Deadline</h5>
+              </Form.Label>
+              <DatePicker
+                className="form-control form-control-lg"
+                onChange={(date): void => setStartDate(date)}
```
charkour, line 110: `onChange()` uses the incorrect setter. This component should it's own `reviewDeadlineDate` state.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+            </Form.Group>
+          </Col>
+        </Row>
+        <Row>
+          <Col md={8}>
+            <Form.Group controlId="description">
+              <Form.Label>
+                <h5>Project Description</h5>
+              </Form.Label>
+              <textarea
```
charkour, line 122: On `textareas`, I think it's nice to set the style `textarea { resize: none; }`.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
```diff
@@ -0,0 +1,297 @@
...
+            if (url === undefined || url === "") {
+              newProjectLinks.delete(linkType);
+            } else {
+              newProjectLinks.set(linkType, url);
+            }
+            setProjectLinks(newProjectLinks);
+            closeLinkModal(linkType);
+          }}
+        />
+      ))}
```
charkour, line 294: Would it be possible to only add one modal, and when it is clicked, you pass the `linkType`, link information, and the setter to update the link? That would allow for fewer components to be in the tree.

ClientApp/src/components/ProjectLinkModal/ProjectLinkModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+              <span>Link to&nbsp;</span>
+              {linkType !== undefined ? linkType.valueOf() : String(linkType)}
+            </Form.Label>
+            <Form.Control
+              defaultValue={formik.values.url}
+              name="url"
+              onChange={formik.handleChange}
+              type="text"
+            />
+            {formik.errors.url}
```
charkour, line 54: Reason this does not using `<Form.Control.Feeback>`? It think it looks nice and integrates Formik well with Bootstrap.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
jonathanchaffer: I agree that would probably be better, and that's what I tried initially, but I couldn't get the modal to re-render correctly when you click on a different project link, so I decided to just make a different modal for each link type. If you checkout commit `f3ddc1b9e402d6be8b407b8c926d2980b1565e01` you'll see what I mean.

ClientApp/src/components/ProjectLinkModal/ProjectLinkModal.tsx
charkour: Wow, I really struggle with English. Sorry about that.

ClientApp/src/components/ProjectEditor/ProjectEditor.tsx
charkour: Oh yeah, this is a trickier problem than initially thought.

charkour: The issue with your previous commit is because you are iterating over the enum values rather than the actual list of of links, so the modal is not getting passed the correct information.

charkour: But I don't know the best way to fix it. It might work to pass the modal information when you click the button to show the modal.

jonathanchaffer: Refactored to use a list of links rather than an enum. Adds the ability to insert custom links as well.

PR #39, opened 2020-06-23
Basic Testing

charkour:
- Provides basic smoke testing to current components
  - In the future, testing the routing and navigation would be good, but we don't have pages to navigate to yet


ClientApp/src/components/App/App.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { render } from "@testing-library/react";
+import { shallow } from "enzyme";
+import React from "react";
+import { BrowserRouter } from "react-router-dom";
+import { App, ReactApp } from "./App";
+
+test("rendering ReactApp without crashing", () => {
+  shallow(<ReactApp />);
+  render(<ReactApp />);
```
mtleonard120, line 11: I don't think this is how the `render` function is typically used. The `render` function _returns_ an object of useful functions for testing various assertions. For example, you could have something like:

```TSX
  const { getByText } = render(<ReactApp />);
  const heading = getByText(/Sign in to Benchmark/i);
  expect(heading).toBeInTheDocument();
```

Note that this is all from a separate testing library altogether called [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) which might be a bit more heavily-featured than what we need. Certainly the example I gave above would not be a "smoke test" kind of test but a more specific test entirely. So you can probably remove this line from the smoke test.

If you're interested though you can find more info about the render function [here](https://testing-library.com/docs/react-testing-library/api#render) and more info about the query functions it returns [here](https://testing-library.com/docs/dom-testing-library/api-queries#variants).

ClientApp/src/components/App/App.test.tsx
```diff
@@ -0,0 +1,22 @@
+/* eslint-env jest */
+
+import { render } from "@testing-library/react";
+import { shallow } from "enzyme";
+import React from "react";
+import { BrowserRouter } from "react-router-dom";
+import { App, ReactApp } from "./App";
+
+test("rendering ReactApp without crashing", () => {
+  shallow(<ReactApp />);
```
mtleonard120, line 10: This line is really all that's needed for a basic smoke test. If rendering fails this function will throw an error failing the test.

Just FYI note that shallow rendering _does not render_ the component's children components, it just sort of places them as if they were standard DOM elements. So in this case we would be rendering precisely:
```TSX
<React.StrictMode>
    <BrowserRouter>
          <App />
    </BrowserRouter>
</React.StrictMode>
```
This means you can do tests saying "I expect there to be exactly 1 App element in the shallow render". This sounds silly in this context but sometimes that's a useful kind of test - like "I expect to see 3 Employee components given x and y props".

ClientApp/src/components/App/App.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+});
+
+test("rendering base App without crashing", () => {
+  const component = (
+    <BrowserRouter>
+      <App />
+    </BrowserRouter>
+  );
+  shallow(<div>{component}</div>);
+  render(<div>{component}</div>);
```
mtleonard120, line 21: Can drop this one.

ClientApp/src/components/App/App.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+  render(<ReactApp />);
+});
+
+test("rendering base App without crashing", () => {
+  const component = (
+    <BrowserRouter>
+      <App />
+    </BrowserRouter>
+  );
+  shallow(<div>{component}</div>);
```
mtleonard120, line 20: I think this line can just be `shallow(<App />);` since otherwise we would not actually be _rendering_ the App component itself.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { shallow } from "enzyme";
+import React from "react";
+import { ConfirmModal } from "./ConfirmModal";
+
+test("rendering ConfirmModal without crashing with basic props", () => {
+  const component = (
+    <ConfirmModal
+      isShowing
+      onCancel={(): void => {}}
+      onConfirm={(): void => {}}
```
mtleonard120, line 14: See the [Jest function mocking docs](https://jestjs.io/docs/en/mock-functions.html) for a better, testable way to implement function props.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+      text="string"
+    />
+  );
+  shallow(<div>{component}</div>);
```
mtleonard120, line 18: I think you'll want to avoid using a <div> wrapper any time you shallow render to prevent your actual component from not being rendered. So just do `<ConfirmModal {...props} />` here where you define `props` above as a constant with type `IConfirmModalProps`.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+  render(<div>{component}</div>);
```
mtleonard120, line 19: All of these can likely be removed from smoke tests.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+test("props on ConfirmModal are correct", () => {
+  const component = (
+    <ConfirmModal
+      isShowing
+      onCancel={(): void => {}}
+      onConfirm={(): void => {}}
+      text="this is my text"
+    />
+  );
+  shallow(<div>{component}</div>);
```
mtleonard120, line 31: Here you can omit the shallow render since that's just duplicating the smoke test.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+      onCancel={(): void => {}}
+      onConfirm={(): void => {}}
+      text="string"
+    />
+  );
+  shallow(<div>{component}</div>);
+  render(<div>{component}</div>);
+});
+
+test("props on ConfirmModal are correct", () => {
```
mtleonard120, line 22: Name here can be something like - `it displays the text prop it is provided`

ClientApp/src/components/App/App.test.tsx
charkour: `<App />` needs to be within `<Router />` or `<BrowswerRouter />` to correctly get history according to [this example](https://testing-library.com/docs/example-react-router).

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
charkour: Ah okay, that makes sense. I was trying to keep from rewriting code but didn't think of that way. Thanks.

ClientApp/src/components/App/App.test.tsx
charkour: Okay thanks, I believe I was getting the React Testing Library `render()` confused with `ReactDOM.Render()` which the [CRA guide suggests](https://create-react-app.dev/docs/running-tests/) as a way to use as a smoke test, so I'll go back to fix that.

ClientApp/src/components/App/App.test.tsx
mtleonard120: Ah good point - in this case shallow rendering is probably not the correct choice then. Doing a react-dom render is a better bet for a smoke test

charkour:
- I have updated tests according to these guidelines:
  - All components are tested with a shallow render and then a normal render to the DOM before being unmounted
  - After that, additional props are tested if applicable

ClientApp/src/components/App/App.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { shallow } from "enzyme";
+import React from "react";
+import ReactDOM from "react-dom";
+import { BrowserRouter } from "react-router-dom";
+import { App, ReactApp } from "./App";
+
+test("rendering ReactApp without crashing", () => {
+  shallow(<ReactApp />);
+  const div = document.createElement("div");
```
mtleonard120, line 11: Setting up the dom element and tearing it down is something you can handle with a `beforeEach()` and a `afterEach()` within your test files. See this [recipe](https://reactjs.org/docs/testing-recipes.html#setup--teardown) and this [jest documentation](https://jestjs.io/docs/en/setup-teardown) for more info on this.

Fine to leave this as is for now but this can help eliminate a lot of unnecessary lines in longer test files.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+    isShowing: false,
+    onCancel: (): void => {},
+    onConfirm: (): void => {},
+    text: "string",
+  };
+  shallow(<ConfirmModal {...props} />);
+  const div = document.createElement("div");
+  ReactDOM.render(<ConfirmModal {...props} />, div);
+  ReactDOM.unmountComponentAtNode(div);
+});
```
mtleonard120, line 37: Generally you don't expect to see smoke tests for particular prop combinations - perhaps a better test here is: "If not showing, the modal does not render any dom elements" and then test that the container is empty.

mtleonard120: Actually looks like you sort of test this on line 66, though I think asserting that _no elements_ are rendered in the dom is a better test.

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+    text: "this is my text",
+    variant: "danger",
+  };
+  const { queryByText } = render(<ConfirmModal {...props} />);
+  expect(queryByText(/this is my cancel text/i)).not.toBeInTheDocument();
+  expect(queryByText(/this is my confirm text/i)).not.toBeInTheDocument();
+});
+test("modal button clicks call correct functions", () => {
+  const cancelSomethingInModal = jest.fn();
+  const confirmSomethingInModal = jest.fn();
```
mtleonard120, line 82: awesome

ClientApp/src/components/ConfirmModal/ConfirmModal.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const props: ConfirmModalProps = {
+    cancelText: "this is my cancel text",
+    confirmText: "this is my confirm text",
+    isShowing: true,
+    onCancel: (): void => cancelSomethingInModal(),
```
mtleonard120, line 87: I _think_ these kind of lines can just be `onCancel: cancelSomethingInModal,`

No need for the extra anonymous function.

ClientApp/src/components/ConfirmModal/ConfirmModal.tsx
```diff
@@ -28,10 +28,19 @@ export const ConfirmModal: React.FC<ConfirmModalProps> = ({
...
+          className="mr-2"
+          data-testid="cancelButton"
+          onClick={onCancel}
+          variant="outline-secondary"
+        >
           {cancelText || "Cancel"}
         </Button>
-        <Button onClick={onConfirm} variant={variant || "primary"}>
+        <Button
+          data-testid="confirmButton"
```
mtleonard120, line 40: This is pretty cool - had not seen these before. Can even have babel strip these out for the prod build, though for our app I don't think that would be necessary.

PR #43, opened 2020-06-24
Merge Story into Develop

charkour:
- Story 11 is completed, so I would like to merge it into develop. 
- This PR contains the file structure updates as well as the project editor. Both have been reviewed in #40 and #33


charkour:
It should be fine, you can change the path if you wish.

PR #52, opened 2020-06-24
Feature/cqlhs 49 User Auth

charkour:
This provides functionality for the user to sign in and be authenticated by the backend.


charkour:
The tests are still a work in progress, but there will be more in the future!

jonathanchaffer:
Can we store the user in cookies so the user stays logged in on refresh? [react-cookies](https://www.npmjs.com/package/react-cookies) should be helpful for this.

ClientApp/src/components/pages/LoginPage/LoginPage.tsx
```diff
@@ -84,7 +73,6 @@ export const LoginPage: React.FC = () => {
           <Form.Group controlId="username">
             <Form.Label>Username</Form.Label>
             <Form.Control
-              disabled={isLoading}
```
jonathanchaffer, line 87: why is this removed?

charkour: Oops, missed that.

charkour:
Yeah, I can add that in another PR after the demo.

PR #60, opened 2020-06-25
changes to db schema +  added employee information + converted to reu…

ekm6:
…sable function. TODO: convert project insert to raw sql


ekm6:
wrong branch

PR #62, opened 2020-06-26
Cookies & logout

jonathanchaffer:
- User stored in cookies on login
- Cookie lifespan is currently set to 1 hour, but this can be changed easily
- Logout button now deletes `user` cookie and sets `context.user` to `undefined`


ClientApp/.eslintrc.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
     "@typescript-eslint",
     "typescript-sort-keys",
     "react-hooks",
+    "sort-keys-fix",
```
charkour, line 33: 🙏 thanks

PR #65, opened 2020-06-29
CQLHS-151 and CQLHS-95 | Project Editor UI and Saving Changes to Backend

jonathanchaffer:
Pretty big PR since it's a combination of CQLHS-151 and CQLHS-95. Major changes are listed here:
- The UI for the `/projects/new` page has been updated to reflect design comps, which are complete according to Ally
- Calendar icon inside datepickers may be tricky to implement, so we decided that's something we could focus on later if we have time
- Form has data validation, including verification that every team member has an employee and role selected
- HTTP request body matches swagger specification
- `apiInstance` function added to initialize any subclass of `BaseAPI` with the right configuration
- PR #70 (Generic AutoComplete component) was merged into this PR, so don't worry about reviewing that stuff since it's already approved


charkour:
Is CQLHS-23 ready to be merged into `develop`?

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -120,11 +122,27 @@ export const ProjectEditPage: React.FC = () => {
                 </Button>
               </Col>
               <Col xs="auto">
-                <Button variant="dark">Save</Button>
+                <OverlayTrigger
+                  // eslint-disable-next-line
+                  overlay={(props) => (
+                    // eslint-disable-next-line
+                    <Tooltip id="save-tooltip" {...props}>
```
charkour, line 129: We should decide if we want to allow prop spreading. It is dangerous because you don't know what fields are on the props object. If we don't want to allow prop spreading, you would then have to write `const { prop1, props2, ... } = props;` but if we do want to allow the prop spreading, then props should be typed so we know what fields are expected on props.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -149,17 +170,94 @@ export const ProjectEditPage: React.FC = () => {
...
+                {errors.client}
+              </Form.Control.Feedback>
+            </Form.Group>
+          </Col>
+        </Row>
+        <Row>
+          <Col xs={4}>
+            <Form.Group className="d-flex flex-column" controlId="start-date">
+              <Form.Label>
+                <small>Start Date</small>
```
charkour, line 195: Date fields need a calendar icon image to match design comps.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
charkour: Also, the line above is return type `JSX.Element`.

jonathanchaffer:
On branch CQLHS-95 I had split up the component into smaller components, so I merged it into this branch. This PR will include changes from both.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
jonathanchaffer: We decided this is probably more work than it's worth, marking as resolved.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
jonathanchaffer: Return type of the function is now `JSX.Element` and the type of `props` is now `OverlayInjectedProps`.

jonathanchaffer:

ClientApp/src/components/App/App.tsx
```diff
@@ -68,10 +69,11 @@ export const App: React.FC = () => {
           <Container>
             <Switch>
               <Route component={LoginPage} path="/login" />
-              <Route component={ProjectEditPage} path="/new-project" />
+              <Route component={ProjectEditPage} exact path="/projects/new" />
```
mtleonard120, line 72: If you place the base routes like `"/projects"` above the nested routes like `"projects/new"` in this Switch component then you don't need the `exact` param. No need to change it, just an fyi

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -68,375 +111,288 @@ export const ProjectEditPage: React.FC = () => {
...
-  };
-  const closeLinkModal = (id: number): void => {
-    setLinkModalsShowing(
-      new Map<number, boolean>(linkModalsShowing.set(id, false)),
-    );
-  };
-
   const history = useHistory();
+  const location = useLocation();
+  const [isNewProject] = useState(location.pathname === "/projects/new");
```
mtleonard120, line 133: `useState` without a state updater function is generally a code smell. Since this is just a constant then it can live above the return statement as just a normal `const` var declaration.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -68,375 +111,288 @@ export const ProjectEditPage: React.FC = () => {
...
+                            onChange={(val): void =>
+                              setFieldValue("startDate", val)
+                            }
+                            placeholderText="MM/DD/YYYY"
+                            value={
+                              values.startDate
+                                ? moment(values.startDate).format("MM/DD/YYYY")
+                                : ""
+                            }
+                          />
```
mtleonard120, line 212: Could you add a `autoComplete="off"` on the date controls? Otherwise you can get some annoying UI like this 
image

ClientApp/src/components/App/App.tsx
jonathanchaffer: That seems like a better design so I tried updating it like this:
`<Route component={ProjectDirectoryPage} path="/projects" />`
`<Route component={ProjectEditPage} path="/projects/new" />`
but that causes `/projects/new` to render `ProjectDirectoryPage`. Did I miss something?

mtleonard120: Turns out I'm a fool and misremembered react-router's default matching algorithm. Without the `exact` prop then RR does partial route matching by default, which means `/projects/new` does in fact match `/projects`. 

So I _think_ you can slap one `exact` prop on `/projects` and then any subpage routes would not need the prop. Sorry for the misdirection on that one! Partial matching is dumb in my opinion :P

PR #67, opened 2020-06-30
new controllers, repos, services, and swagger documentation

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/ProjectReviewsController.cs
```diff
@@ -10,6 +10,7 @@ namespace Benchmark.WebApi.Controllers
 {
     [Route("api/[controller]")]
     [ApiController]
+    [Authorize]
```
joeldevries, line 13: Should we be checking certain roles on the action methods in this controller?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/EmployeesController.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Benchmark.WebApi.Controllers
 {
-    //[Authorize]
+    [Authorize]
```
joeldevries, line 11: Should we be checking certain roles on the action methods in this controller?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/ProjectsController.cs
```diff
@@ -19,44 +19,120 @@ public ProjectsController(IProjectManagementService projectManagementService)
...
+        ///
+        /// </remarks>
+        /// <returns>A list of projects mapped to the class "projectOverview" class.</returns>
+        /// <response code="200">Returns the items</response>
+        /// /// <response code="404">Project Manager not found</response>
+        /// <response code="401">Unauthorized</response>
         [Authorize(Roles = "PM")]
-        [HttpGet("GetManagedProjects")]
+        [HttpGet("ManagedProjects{id}")]
         public async Task<ActionResult<IList<Employee>>> GetManagedProjects(int id)
```
joeldevries, line 38: I noticed that lots of places in the API return the exact object that is mapped from the database. We often create a different model to pass over the web so that we don't expose internal data to places it shouldn't go. Something we should consider going forward.

tonysulfaro: Agreed. some of the branches that are ahead of this branch (71,72) will resolve this

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/EmployeesController.cs
tonysulfaro: We haven't gotten that far yet with permissions with regard to employees. After meeting with ally though it would appear that everyone that is authenticated with the app would have access.

PR #69, opened 2020-06-30
feature/CQLHS-82 | Project Details page UI

jonathanchaffer:
- Uses a `placeholderData` object to populate the page.
- The ProjectDetailsPage component is long; needs to be split into smaller components once it uses actual data.
- Also made a minor update to readme.


ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+        <Col md={4} xs={12}>
+          <Row>
+            <Col>
+              <h5>Additional Information</h5>
+            </Col>
+          </Row>
+          <Row>
+            <Col>
+              <span>
+                Project Date Span:{" "}
```
charkour, line 179: Missing the last modified section that are on design comps (a recent addition), which should be below a "Created on MM/DD/YYYY by First Lastname" section.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+              >
+                Delete
+              </Button>
+            </Col>
+          </Form.Row>
+        </Col>
+      </Row>
+      <Form.Row>
+        <Col>
+          <h5>
```
charkour, line 88: Why do you omit `<h1>`, and then skip from `<h2>` to `<h5>`? It seems you would want to start with `<h1>` or `<h2>` (maybe) and then increment from there. "The Team" and "Additional Info" should probably be the same font-weight (or more) as the the "Links" and "Notes" sub-headings under "Adtl Info"

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+              </span>
+            </Col>
+          </Row>
+          <Row>
+            <Col>
+              <h6>Links</h6>
+            </Col>
+          </Row>
+          <Row>
+            <Col className="d-flex flex-wrap">
```
charkour, line 197: The wrapping looks nice.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+  const {
+    context: { user },
+  } = useContext(AppContext);
+  const history = useHistory();
+  const location = useLocation();
+  return (
+    <div className="ProjectDetailPage">
+      <Row>
+        <Col>
+          <BackButton text="Back" />
```
charkour, line 58: Back button missing functionality.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+        onConfirm={(): void => setIsShowingConfirmDelete(false)}
+        text="Are you sure you want to delete this project? Team members will no longer have access to the project evaluation survey."
+        variant="danger"
+      />
+      <ConfirmModal
+        confirmText="Send Reminder"
+        isShowing={isShowingConfirmSendReminder}
+        onCancel={(): void => setIsShowingConfirmSendReminder(false)}
+        onConfirm={(): void => setIsShowingConfirmSendReminder(false)}
+        text="Are you sure you want to send an email reminder to this user?"
```
charkour, line 234: Again, I think we should make a generic modal so we only need one per page, but that would be for another subtask!

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
jonathanchaffer: Updated the headings to be consistent with comps. As long as those are consistent, we can freely update global heading styles to match what is specified in comps, and not worry about styling each individual heading.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
jonathanchaffer: Should it go to the previous page regardless of where they came from, or should it always go to the Project Directory?

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
jonathanchaffer: I think it makes sense having these as separate modals since they are confirmations of different things. Hypothetically if we wanted to pull up a confirmation modal from within a confirmation modal, this would allow us to do it.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.scss
```diff
@@ -0,0 +1,12 @@
+.ProjectDetailPage {
+  .row {
+    margin-bottom: 1em;
+  }
+
+  .table th,
+  .table td {
+    padding: 0.25rem;
+    vertical-align: middle;
+    height: 40px;
```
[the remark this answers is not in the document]
jonathanchaffer: Agreed! We should make as many styles global as possible so we don't have to style individual components.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
charkour: Hmmm, that is a good question. The browser has a back button so maybe that would be "Back to Project Directory" or something along those lines. We should probably ask Ally what the UX for that is.

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
charkour: Ohh okay that makes sense. Didn't thank of that.

jonathanchaffer:
Now building the reusable header component, which includes the back button. Once that's approved, I'll update the back button here.

jonathanchaffer:

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+                        <th>
+                          <h6>Role</h6>
+                        </th>
+                        {user?.accessLevel === "admin" ? (
+                          <>
+                            <th>
+                              <h6>Review Status</h6>
+                            </th>
+                            <th className="d-none">
+                              <h6>Notify</h6>
```
mtleonard120, line 82: In the comps it looks like this reads as "Action". Also why the `d-none`?

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const [
+    isShowingConfirmSendReminder,
+    setIsShowingConfirmSendReminder,
+  ] = useState(false);
+  const {
+    context: { user },
+  } = useContext(AppContext);
+  return (
+    <div className="ProjectDetailPage">
+      {/* TODO: use reusable Header component */}
```
mtleonard120, line 55: Until this component is available I'm not sure I would really call this subtask done (or else we would need to add a "Incorporate header component" subtask on this story).

jonathanchaffer:
todo: add back button to header

jonathanchaffer:
todo: update notify button

jonathanchaffer:
todo: highlight items on hover

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -63,72 +88,70 @@ export const ProjectDetailPage: React.FC = () => {
           <Row>
             <Col>
               <Card>
-                <Card.Body>
-                  <Table borderless className="mb-0">
```
mtleonard120: I'm a little concerned that this is no longer a Table, which would be more accurate semantically. Not a huge deal for an internal app but if this was public facing, screen readers and search crawlers alike are going to have a much easier time understanding tabular data than just a grid of divs.

Down the road if this table pattern is reused on other pages you should consider extract a reusable Table component that accepts an array of header strings/elements and then an array of data rows to display and their respective click handlers. Definitely beats writing out all these elements in every place you need it.

jonathanchaffer: Fair point. Using a `Table` was weird on the Project Directory page (didn't render shadows correctly in Safari) so I changed it to use `Row`s and `Col`s, and wanted to be consistent here. A reusable component would definitely be helpful.

PR #71, opened 2020-07-01
feature/CQLHS-61 | Project Evaluations Table

charkour:
This PR merges the Project Evaluations Table page into the story branch.
- Add a large table to see the projects separated by year.
- Clicking on a table row will navigate to the Project Evaluation Over page or Editor based on the status.
- There is test data until the backend finishes the API, but that is completed in a different task
- Searching and filtering is completed in another task
- Added custom card and shadow classes. I don't know all bootstrap styles, so let me know if I should use them.

Bug: 
- The sticky table header does not cover the table shadows which does not look good. 
- I do not know how to fix this, so please help with that.

Let me know what I missed and need to fix so it works better! Specifically in terms of SCSS. I am better at TS than styling.


charkour:
There are two warnings about having `<div>`s and `<h4>` in `<tables>` but it seems to be working fine. Maybe those can be ignored?

jonathanchaffer:
Cleaned up the table styles a bit to reduce the need to remember which classes should be applied to each table element.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectTableGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+}: FinishButtonProps) => {
+  return (
+    <td className="FinishButton d-flex justify-content-between align-items-center">
+      <span>{projectStatus}</span>
+      <Button size="sm" variant="primary">
+        Finish
+      </Button>
+    </td>
+  );
+};
```
jonathanchaffer, line 81: This also is small enough that it probably doesn't need to be its own component.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -1,12 +1,74 @@
...
+    <div className="th">Start and End Date</div>
+    <div className="th">Evaluation Status</div>
   </div>
 );
+
+export const TableBody: React.FC = () => (
+  <>
+    {testData.map((group, idx) => (
+      <>
+        <h4
```
jonathanchaffer, line 65: Can the year text be aligned with the table contents?

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -1,12 +1,69 @@
...
+        <Col xs="auto">
+          <Link to="/projects/new">
+            <Button>New Project Evaluation</Button>
+          </Link>
+        </Col>
+      </Row>
+      <hr className="divider" />
+      <Row>
+        <Col className="d-flex flex-row-reverse mt-2">
+          <ProjectSearchBar />
```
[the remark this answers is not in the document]
charkour: Sticky search would probably be good!

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -1,12 +1,74 @@
...
+        <h4
+          className={`groupTitle ${idx !== 0 ? "mt-4" : ""}`}
+        >{`Year ${group.year}`}</h4>
+        <tbody key={group.id} className="widget">
+          <ProjectTableGroup key={group.id} group={group} />
+        </tbody>
+      </>
+    ))}
+  </>
+);
```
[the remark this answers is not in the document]
charkour: Okay, I just thought they looked ugly in the main the ProjectDirectoryPage FC and knowing that the Page has a `<Table>` with a `<Head>` and `<Body>` is a clean abstraction. You don't really care how they are implemented on the page. I agree that they are pretty small, so that's why I liked breaking them into little FCs but kept it in the same file. I can put them in their own files if you want, but I didn't want to bloat the repo with files. Should I put them in their own files?

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectTableGroup.tsx
charkour: This logic makes the parent component's JSX tree messy so it is nice to have it abstracted.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
charkour: Could you elaborate on what you are suggesting? Currently, the year text is aligned with the table and matches the design comps. Having the content of column one be right-offset of the year text makes sure that year does not look like a column heading.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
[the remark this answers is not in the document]
jonathanchaffer: Yeah the abstraction makes sense. If they're separate they should probably be in their own files. Maybe rename to `ProjectDirectoryTableHead` and `ProjectDirectoryTableBody`.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
jonathanchaffer: Maybe it's just the text size that makes it look weird. Once global styles match design comps it'll probably look better.

ClientApp/src/styles/index.scss
```diff
@@ -19,3 +19,25 @@ html {
...
+  margin-top: 0.25em;
+  margin-bottom: 0.25em;
+  border-width: 1.5px;
+}
+
+// Override default Bootstrap shadow to match design comps
+.shadow {
+  -webkit-box-shadow: 0px 2px 4px 0px rgba(0, 0, 0, 0.2) !important;
+  -moz-box-shadow: 0px 2px 4px 0px rgba(0, 0, 0, 0.2) !important;
+  box-shadow: 0px 2px 4px 0px rgba(0, 0, 0, 0.2) !important;
```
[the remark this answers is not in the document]
charkour: If you have suggestions on what we should do for that, please let me know or make the changes after!

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -1,12 +1,74 @@
...
+          <ProjectSearchBar />
+        </Col>
+      </Row>
+      <Row>
+        <Col>
+          <Table
+            borderless={true}
+            className="ProjectTable mb-5 mt-4"
+            hover={true}
+            striped={true}
```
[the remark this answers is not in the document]
charkour: Update: no stripes.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
charkour: Sounds good. And I think it'll be changed again to match the designs. No worries.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectTableGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+interface ProjectTableGroupProps {
+  group: TestProjectDataGroup;
+}
+
+/*
+  Displays the list of projects for a specific year
+  Row is clickable and redirects to Project Overview (or Editor if a draft)
+*/
+export const ProjectTableGroup: React.FC<ProjectTableGroupProps> = ({
```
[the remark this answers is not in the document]
charkour: This is rendering multiple rows through the `map` function.

charkour:
Updates: 
- Updates the table's to match more of the design standards.
- Some styles on the search bar need to be fixed. 
- It will now hide columns on mobile
- Shadows are updated to a variable
- Some code refactored

jonathanchaffer:

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <Col>Status</Col>
+        <Col />
+      </Row>
+      {data.map((group) => (
+        <>
+          <Row>
+            <Col>
+              <h4>{group.year}</h4>
+              <Card>
+                <ListGroup>
```
jonathanchaffer, line 39: Apply `variant="flush"` to `ListGroup`

jonathanchaffer:
todo: highlight items on hover

jonathanchaffer:
Removed the Finish button since Ally and I decided that saving a project as a draft was nonessential. So the project evaluations table only shows project evals that have been sent out. Users can still click into a project eval and edit its details.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+import React, { useContext } from "react";
+import Card from "react-bootstrap/Card";
+import Col from "react-bootstrap/Col";
+import ListGroup from "react-bootstrap/ListGroup";
+import Row from "react-bootstrap/Row";
+import { useHistory } from "react-router-dom";
+import { TestProjectData, TestProjectDataGroup } from "./TestData";
+
+interface ProjectDirectoryTableProps {
+  data: TestProjectDataGroup[];
```
mtleonard120, line 12: I do recommend using more descriptive prop names whenever possible. `projectGroups` might be a better choice here.

ClientApp/src/components/reuseables/Header/BackButton/BackButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  path,
+}: BackButtonProps) => {
+  const history = useHistory();
+  return (
+    <div className="BackButton">
+      <Button
+        className="text-secondary d-flex align-items-center"
+        onClick={(): void => {
+          if (!path) history.goBack();
+          else history.push(path);
```
mtleonard120, line 23: Just a warning that these sorts of manual stack manipulations can get buggy pretty quick so make sure you tinker with them to make sure the UX feels right. For instance you might consider using `history.replace` here instead to remove the current page from the stack

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
jonathanchaffer: Did some refactoring to use a list of actual Project objects. The fields aren't all there yet so there are some placeholders, but once the API is updated we can plug in those values.

ClientApp/src/components/reuseables/Header/BackButton/BackButton.tsx
jonathanchaffer: Good point, but I think `history.push` makes sense for this component, since you might want to "undo" going back by hitting the browser's back button. If you use `history.replace` then you can't get back to the page you were on unless you retype the URL. I haven't looked too much into the other `history` functions available, so let me know if there are other options I should consider.

PR #72, opened 2020-07-01
get updated migrations to modify project CRUD methods

ekm6:


ekm6:
jus need files, no need to merge all these files

PR #77, opened 2020-07-01
feature/CQLHS-156 | Reusable Header component and global styles

jonathanchaffer:
- `actions` prop is an array of `JSX.Element`s (usually `Button`s). The `Header` component automatically spaces and aligns them to the upper right.
- `backButtonText` specifies the text of the back button; `backButtonPath` specifies the path it redirects to. If no `backButtonText` is provided, defaults to `Back`. If no `backButtonPath` is provided, defaults to the previously visited page. If neither are provided, back button is not rendered.
- You can pass a `string` or `JSX.Element` as the `subtitle` prop in case you want parts of it to be bold or italic, e.g. `subtitle={<span>Client: <strong>Petsmart</strong></span>}`
- Also added a `StylePreviewPage` as an easy way for to verify that styles match comps.
- `.row` has `margin-bottom: 1em` globally, but this can be adjusted at the component level when needed.

This PR merges `feature/CQLHS-156` into `story/CQLHS-20`, but maybe it should be merged into `develop` instead, so the global styles can be used on other stories. What do you think @mtleonard120 ?


ClientApp/src/components/pages/StylePreviewPage/StylePreviewPage.tsx
```diff
@@ -0,0 +1,64 @@
+import { Header } from "components";
+import React from "react";
+import Button from "react-bootstrap/Button";
+import Col from "react-bootstrap/Col";
+import Row from "react-bootstrap/Row";
+
+export const StylePreviewPage: React.FC = () => {
```
mtleonard120, line 7: I dig this - sort of like `storybook` but just for styles. It's arguable whether this page should be accessible on production builds. Unless you anticipate Ally needing to look at this on the dev site you could wrap this page in a `NODE_ENV` check so that it just redirects to the dashboard or something when we're in a production build.

That would look like
```TSX
if (process.env.NODE_ENV === "production") {
  return <Redirect to="/dashboard" />
} else {
  return ...
}
```
I'll leave it up to your best judgement.

jonathanchaffer: I can see it being useful for Ally so I'll keep it accessible for now.

PR #85, opened 2020-07-02
feature/CQLHS-162 | Navbar logo & responsive layout

jonathanchaffer:
Conditionally renders links based on user `accessLevel`.


ClientApp/src/components/Navigation/Navigation.tsx
```diff
@@ -17,44 +19,57 @@ export const Navigation: React.FC = () => {
...
+                  )}
+                  {user.accessLevel === "admin" || user.accessLevel === "pm" ? (
+                    <Link to="/projects">
+                      <Button variant="link">
+                        <span>Project Evaluations</span>
+                      </Button>
+                    </Link>
+                  ) : (
+                    <></>
+                  )}
```
mtleonard120, line 59: This is fine as it is, but a reminder that if you ever have repeated JSX in a component you can likely extract a helper component. Doing this keeps your JSX more declarative and especially pays off if you end up needing something similar anywhere else in the app.

In this case you could have a reusable like a `BenchmarkLink` component with the following API:

```TSX
<BenchmarkLink label="Dashboard" limitedTo={["admin", "teamLead"]} to="/dashboard"/>
<BenchmarkLink label="Employee Directory" limitedTo={["admin"]} to="/employees"/>
<BenchmarkLink label="Project Evaluations" limitedTo={["admin", "pm"]} to="/projects" />
```

Consuming the `AppContext` could occur inside the implementation of this component. So for now keep this as it is but be on the lookout moving forward for helper component opportunities like this.

PR #88, opened 2020-07-06
CQLHS-166 | Project Evaluations Table Header Not Sticky

jonathanchaffer:
Header no longer sticky; also added a flag to determine whether the table should be split by year or not, which will be used when sorting by alphabetical attributes.


ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
```diff
@@ -18,48 +18,90 @@ interface ProjectDirectoryTableProps {
 export const ProjectDirectoryTable: React.FC<ProjectDirectoryTableProps> = ({
   projects,
 }: ProjectDirectoryTableProps) => {
-  const {
-    context: { user },
-  } = useContext(AppContext);
+  const [isSplitByYear, setIsSplitByYear] = useState(true);
```
mtleonard120, line 21: As-is I don't see any logic here to update this state variable. Either there should be a `useEffect` to set this value based on `props.projects` or you should add a comment indicating how this will be handled in an upcoming subtask.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
```diff
@@ -18,48 +18,90 @@ interface ProjectDirectoryTableProps {
...
+      )}
     </>
   );
 };
 
-interface ProjectTableGroupProps {
+export const ProjectTableHeader: React.FC = () => {
+  const {
+    context: { user },
+  } = useContext(AppContext);
```
mtleonard120, line 65: I feel like I've seen this a lot in various areas. Any time you have repeated hook logic it's a good sign you should consider making a custom hook. In this case, you could add a custom hook to the effect of:

```TSX
// the somewhat silly name here is driven by custom hook convention
export const useUser = () => {
  const {
    context: { user }
  } = useContext(AppContext);
  
  return user;
}
```
Then anytime you need the `user` you can just do
```TSX
const user = useUser();
```
This custom hook could likely live in `authService` or even a new `contextService` where other similar helper hooks based on contexts could eventually live.

PR #89, opened 2020-07-06
fixed scaffolding issues, some work on fixing bug with updating links

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/EmployeesController.cs
```diff
@@ -55,7 +57,7 @@ public async Task<ActionResult<IList<Employee>>> GetAllEmployees()
         /// <response code="401">Unauthorized</response>
         /// <response code="404">Not Found</response>
         [HttpGet("{id}")]
-        public async Task<ActionResult<Employee>> GetEmployee(int id)
```
joeldevries, line 58: This is much better than previous - I was going to say that it's not a great idea to return the objects from the DbContext directly to the web, but I see you are doing that already.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/Employee.cs
```diff
@@ -19,6 +21,37 @@ public Employee()
             RefreshTokens = new HashSet<RefreshToken>();
         }
 
+        public Employee(EmployeePostRequest request)
```
joeldevries, line 24: I think it's better to do this mapping type code outside of the Employee object. Also, check out AutoMapper - it can really save time for stuff like this.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Response/EmployeeOverview.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Benchmark.WebApi.Models.Database;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Benchmark.WebApi.Models.Request
```
joeldevries, line 3: This should be `Benchmark.WebApi.Models.Response`

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ClientRepository.cs
```diff
@@ -31,7 +28,10 @@ public async Task<OperationResult<Client>> GetClientFromName(string clientName)
                 .FirstOrDefaultAsync();
 
             if (client == null)
+            {
```
joeldevries, line 31: Make sure your Client table has a unique index on ClientName if you want to lookup by name, otherwise you could get duplicates and apply changes to the wrong record.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/EmployeeRepository.cs
```diff
@@ -37,14 +38,19 @@ public async Task<OperationResult<Employee>> GetEmployeeFromName(string[] employ
                 .Where(e => e.FirstName == employeeName[0] && e.LastName == employeeName[1])
                 .FirstOrDefaultAsync();
             if (employee == null)
```
joeldevries, line 40: This is a weird thing to do pass name as a string array. `(string firstName, string lastName)` would be better. Also, not sure this is good, unless you have a unique index on FirstName, LastName, and I'm not sure that is good either.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -116,7 +116,7 @@ public async Task<OperationResult> DeleteProject(int id)
             return OperationResult.Ok();
         }
 
-        public async Task<OperationResult> PostProject(Project project)
+        public async Task<OperationResult<PostResponse>> PostProject(Project project)
```
joeldevries, line 119: If I'm not mistaken Post is supposed to update and Put is supposed to create new records? 

Regardless, the data layer, IMHO should not differentiate between them. You could either have a Save method that determines whether to create or update, or you could have a Create and Update method.

Having a Put and a Post on the repository is not clear on what you are doing.

Also - does that PutProject method actually work? I have not seen EF used in that way.

PR #91, opened 2020-07-06
CQLHS-79 | Sort Project Table Columns

jonathanchaffer:
- Add the ability to click on a column to sort by that attribute. Click again to reverse.

PR #92, opened 2020-07-06
CQLHS-81 | Filter Projects by Search Term

jonathanchaffer:
- Search works with project title, PM, client, and start/end year.


jonathanchaffer:
If #91 is merged before this there might be merge conflicts, so let me know if I need to deal with those.

ClientApp/src/services/projectService.ts
```diff
@@ -16,3 +16,28 @@ export const groupProjectsByYear = (
...
   return groups;
 };
+
+export const filterProjectsByTerm = (
+  projects: Project[],
+  term: string,
+): Project[] => {
+  const standardize = (str: string): string => {
+    return str.toLowerCase().trim();
+  };
```
mtleonard120, line 26: If this function ends up being a common need across services you can add a `utils` directory and plop this in a `stringUtils` file.

PR #102, opened 2020-07-08
CQLHS-174 | update nav link styles

jonathanchaffer:


ClientApp/src/components/Navigation/Navigation.tsx
```diff
@@ -32,7 +33,12 @@ export const Navigation: React.FC = () => {
                   {user.accessLevel === "admin" ||
                   user.accessLevel === "teamLead" ? (
                     <Link to="/dashboard">
-                      <Button variant="link">
+                      <Button
+                        className={
+                          location.pathname === "/dashboard" ? "current" : ""
+                        }
+                        variant="link"
+                      >
```
mtleonard120, line 41: This logic is repeated a few times in the file, which calls for a helper component to be used. 

In this case, make your own `NavButton` component is this file that accepts a single prop: `activeRoute: string` that can be used for the various. Then if we need to adjust all of these instances with any new logic or classes we only update a single helper component.

PR #112, opened 2020-07-09
Story/cqlhs 24

ekm6:
there are front end conlflicts as well


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ClientRepository.cs
```diff
@@ -49,7 +49,8 @@ public async Task<OperationResult<int>> PostClient(Client client)
                 throw;
             }
 
-            return OperationResult.Ok(GetClientFromName(client.Name).Result.Data.ClientId);
+            OperationResult<Client> response = await GetClientFromName(client.Name);
```
joeldevries, line 52: You don't need to look up the client again. EF will automatically update the ClientId when creating a new one.

So you can just do this:
```
await context.SaveChangesAsync();
return OperationResult.Ok(client.ClientId);
```

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/EmployeeRepository.cs
```diff
@@ -115,40 +77,40 @@ public async Task<OperationResult> PostRefreshToken(Employee employee, RefreshTo
...
         }
 
-        public async Task<OperationResult> PutEmployee(int id, Employee employee)
+        public async Task<OperationResult<PostResponse>> UpdateEmployee(Employee employee)
         {
             await using BenchmarkContext context = new BenchmarkContext(_options);
-            context.Entry(employee).State = EntityState.Modified;
+            if (employee.EmployeeId > 0)
+            {
+                await context.ReconcileAsync(employee, e =>
```
joeldevries, line 85: Do you plan to continue to use Reconcile?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectEvaluationsRepository.cs
```diff
@@ -22,62 +22,62 @@ public ProjectEvaluationRepository(ILogger<ActiveDirectoryRepository> logger, Db
...
-            await context.ProjectEvaluation.AddAsync(evaluation);
+            await context.ProjectEvaluationResult.AddAsync(evaluation);
             try
             {
                 await context.SaveChangesAsync();
             }
             catch (DbUpdateException ex)
             {
-                if (ProjectEvaluationExists(evaluation.ProjectId).Result.Data)
+                if (await ProjectEvaluationExists(evaluation.ProjectId) == OperationResult.Ok(true))
```
joeldevries, line 80: I don't think this comparison will work since you are comparing two OperationResults when you should probably be checking the Data property of ProjectEvaluationExists(evaluation.ProjectId)

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectEvaluationsRepository.cs
```diff
@@ -102,7 +102,7 @@ public async Task<OperationResult> PutEvaluation(int id, ProjectEvaluation evalu
             }
             catch (DbUpdateConcurrencyException ex)
             {
-                if (!ProjectEvaluationExists(id).Result.Data)
+                if (await ProjectEvaluationExists(id) == OperationResult.Ok(true))
```
joeldevries, line 105: Same as above - I think you had it right in the first version.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -55,47 +50,41 @@ public async Task<OperationResult<IList<ProjectOverview>>> GetAllProjects()
...
-                {
-                    ProjectId = p.ProjectId,
-                    ClientName = p.Client.Name,
-                    ManagerName = p.Manager.FirstName + ' ' + p.Manager.LastName,
-                    ProjectName = p.ProjectName,
-                    Links = p.Link.Select(l => new LinkResponse(l)),
-                    StartDate = p.StartDate,
-                    EndDate = p.EndDate,
-                })
+                .Include(p => p.ProjectEvaluationResult)
```
joeldevries, line 68: This is not a very efficient way to do this - this will run at least 6 queries and potentially many, many more on the database. This would be a big no-no in a high-traffic system, as query overhead can become a big deal. Probably not a problem in this system, so don't bother changing it.

This is true anytime you use .Include and even more true when you use ThenInclude.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -55,47 +50,41 @@ public async Task<OperationResult<IList<ProjectOverview>>> GetAllProjects()
...
+                    .ThenInclude(per => per.Role)
+                .Include(p => p.ProjectEvaluationResult)
+                    .ThenInclude(per => per.ReviewerEmployee)
+                    .ThenInclude(e => e.EmployeeRole)
+                    .ThenInclude(er => er.Role)
+                    .ThenInclude(r => r.Department)
+                .Select(p => new ProjectDetails(p))
                 .FirstOrDefaultAsync();
+            return (project != null) ? OperationResult.Ok(GetEvaluationPercentage(project)) : OperationResult.NotFound<ProjectDetails>($"Project with id{id} not found");
```
joeldevries, line 77: Check out the OperationResult.FromValue method - looks to me like it does the exact same thing, but your code will look a lot neater (easier to read).

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -116,53 +105,43 @@ public async Task<OperationResult> DeleteProject(int id)
...
         {
             await using BenchmarkContext context = new BenchmarkContext(_options);
-            await context.Project.AddAsync(project);
-            try
-            {
-                await context.SaveChangesAsync();
-            }
-            catch (DbUpdateException ex)
+
+            var client = context.Client.Local.FirstOrDefault(cl => cl.ClientId == request.Client.ClientId) ??
```
joeldevries, line 112: What is the point of using context.Client.Local instead of context.Client?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -116,53 +105,43 @@ public async Task<OperationResult> DeleteProject(int id)
...
+                         new Client(request.Client);
+
+            context.Client.Attach(client);
```
joeldevries, line 115: Why do you have to Attach the client? Is this a side-effect of using context.Client.Local?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -116,53 +105,43 @@ public async Task<OperationResult> DeleteProject(int id)
...
+                Client = client
+            };
 
-        public async Task<OperationResult> PutProject(int id, Project project, ICollection<Link> newlinks)
-        {
-            await using BenchmarkContext context = new BenchmarkContext(_options);
-            context.Entry(project).State = EntityState.Modified;
+            foreach (var eval in newProject.ProjectEvaluationResult)
+            {
+                eval.Role.Department = deptObjs[eval.Role.DepartmentId];
```
joeldevries, line 131: You are changing the Department of a Role here. Are you sure you want to do that? (correct answer is No).

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Repositories/ProjectRepository.cs
```diff
@@ -116,53 +105,43 @@ public async Task<OperationResult> DeleteProject(int id)
...
+            }
+            await context.AddAsync(newProject);
```
joeldevries, line 133: This method is called UpdateProject, but here we are Creating a new one. The method name should reflect closely what the method is doing.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/AuthService.cs
```diff
@@ -99,8 +99,8 @@ public async Task<OperationResult<Employee>> GetEmployeeFromAccessToken(string a
                 if (jwtSecurityToken != null && jwtSecurityToken.Header.Alg.Equals(SecurityAlgorithms.HmacSha256, StringComparison.InvariantCultureIgnoreCase))
                 {
                     string employeeId = principle.FindFirst(ClaimTypes.Name)?.Value;
-                    EmployeeOverview employee = _employeeRepository.GetEmployee(Convert.ToInt32(employeeId)).Result.Data;
-                    return OperationResult.Ok(new Employee(employee));
+                    OperationResult<EmployeeOverview> employee = await _employeeRepository.GetEmployee(Convert.ToInt32(employeeId));
```
joeldevries, line 102: Convert.ToInt32 will throw an exception if the conversion fails. I would use int.TryParse and handle the failure gracefully.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmployeeService.cs
```diff
@@ -29,21 +29,21 @@ public async Task<OperationResult<EmployeeOverview>> GetEmployee(int id)
             return await _employeeRepository.GetEmployee(id);
         }
 
-        public async Task<OperationResult> PutEmployee(int id, EmployeePutRequest request)
+        public async Task<OperationResult> PutEmployee(int id, EmployeeUpdateRequest request)
```
joeldevries, line 32: We don't need the id parameter and it is redundant - we already have the EmployeeId as part of the EmployeeUpdateRequest object.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmployeeService.cs
```diff
@@ -29,21 +29,21 @@ public async Task<OperationResult<EmployeeOverview>> GetEmployee(int id)
...
             {
                 return OperationResult.Cancelled("Bad request");
             }
 
             Employee employee = new Employee(request);
-            return await _employeeRepository.PutEmployee(id, employee);
+            return await _employeeRepository.UpdateEmployee(employee);
         }
 
-        public async Task<OperationResult<PostResponse>> PostEmployee(EmployeePostRequest request)
```
joeldevries, line 43: PutEmployee and PostEmployee do essentially the same thing. If I was a new developer on this project, how would i know which one to use when creating a new employee? Or updating an existing employee? I think one of these could be removed.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/ProjectManagementService.cs
```diff
@@ -40,87 +39,35 @@ public async Task<OperationResult<IList<ProjectOverview>>> GetAllProjects()
...
-            }
-            else
-            {
-                project.Client = new Client(clientname);
-            }
-            return OperationResult.Ok(project);
-        }
-
-        public async Task<OperationResult<PostResponse>> PostProject(ProjectPostRequest request)
+        public async Task<OperationResult> UpdateProject(ProjectUpdateRequest request, int id = -1)
```
joeldevries, line 57: Same redundant id parameter. Please remove.

Benchmark.WebApi/Benchmark.WebApi/BenchmarkTester/ReconcilerTests.cs
```diff
@@ -0,0 +1,168 @@
+﻿using Benchmark.WebApi.Models.Database;
```
joeldevries, line 1: This file can be deleted (or commented out) - it was not intended to be a long-term test.

ClientApp/package.json
```diff
@@ -14,6 +14,9 @@
     "@types/jest": "^24.9.1",
     "@types/node": "^12.12.47",
     "@types/react": "^16.9.36",
+    "@types/react-autocomplete": "^1.8.5",
```
joeldevries, line 17: Is this change intentional?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/ProjectEvaluationResult.cs
```diff
@@ -10,14 +10,28 @@ public ProjectEvaluationResult()
             EvaluationQuestion = new HashSet<EvaluationQuestion>();
         }
 
+        public ProjectEvaluationResult(EvaluationPostRequest request)
+        {
+            Role = new Role()
```
joeldevries, line 15: Are you intending to change the DepartmentId and Title of roles here? Doesn't seem like something you want to do in when updating a project evaluation. If you are not (which I assume you are not), why do we accept these values in the post and apply them to the database?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Request/EvaluationPostRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Benchmark.WebApi.Models.Response;
+
+namespace Benchmark.WebApi.Models.Request
+{
+    public class EvaluationPostRequest
+    {
+        public RoleSnippet Role { get; set; }
+        public EmployeeSnippet Employee { get; set; }
```
joeldevries, line 8: I think this should just be int RoleId and int EmployeeId?

ClientApp/package.json
ekm6: @jonathanchaffer what do you think?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/ProjectEvaluationResult.cs
ekm6: We don't change department, but we add a new entry to the roles table based on the project.

ClientApp/package.json
jonathanchaffer: yup!

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/ProjectEvaluationResult.cs
joeldevries: That sounds like a bad idea. Roles are not project-based. They should be created intentionally and only when a new role is actually needed.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/ProjectsController.cs
```diff
@@ -1,14 +1,15 @@
...
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Benchmark.WebApi.Controllers
 {
-    [Authorize]
+    //[Authorize]
+    [AllowAnonymous]
```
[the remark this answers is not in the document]
ekm6: yes

ClientApp/package.json
ekm6: it was intentional

Benchmark.WebApi/Benchmark.WebApi/Benchmark.DbMigrations/Migrations/2020/07/20200707162033_addProjFields.cs
```diff
@@ -0,0 +1,42 @@
...
+            Create.ForeignKey()
+                .FromTable("ProjectEvaluationResult").ForeignColumn("RoleID")
+                .ToTable("Role").PrimaryColumn("RoleID");
+
+            Alter.Table("ProjectEvaluationResult")
+                .AlterColumn("RoleID").AsInt32().NotNullable();
+        }
+
+        public override void Down()
+        {
```
tonysulfaro, line 39: I noticed we had alot of empty down methods, it might be advantageous to have a way to reverse these in the future

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Request/EvaluationPostRequest.cs
ekm6: we expect the client to modify the title, which is saved on the roles table. And we will use the employeeSnippet parameter "email" to send an email as soon as the project is posted.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.DbMigrations/Migrations/2020/07/20200707162033_addProjFields.cs
ekm6: We had to do Execute.Sql() to add non-nullable fields without dropping all the data not in the table, but we'll try to use auto reversing as aften as we can.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Request/EvaluationPostRequest.cs
joeldevries: I have not seen the UI for this, so I don't know what they are sending, but regardless, we can't change the title of a role when creating a project evaluation. If you do that, the role title might change on each project, which is definitely not expected. If you absolutely need project-specific roles (which I doubt you do), then you need a new ProjectRole table to store them in.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.DbMigrations/Migrations/2020/07/20200708100455_moveDeadlineEvalToProj.cs
```diff
@@ -0,0 +1,19 @@
+using FluentMigrator;
+
+namespace Benchmark.DbMigrations.Migrations._2020._07
+{
+    [Migration(20200708100455)]
+    public class _20200708100455_moveDeadlineEvalToProj : Migration
+    {
+        public override void Up()
+        {
+            Delete.Column("Deadline").FromTable("ProjectEvaluationResult");
```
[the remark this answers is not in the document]
ekm6: We originally had deadline there, but now the deadlines are the same for all evaluations in the project, so we thought it would better to save it once on the project instead of on all the individual evaluations. Would it be advantageous to store it on each evaluation?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/ProjectEvaluationResult.cs
ekm6: makes for an easier update as well

PR #113, opened 2020-07-10
CQLHS-84 | Map retrieved data to Project Details component state

jonathanchaffer:
Sample data on the database is currently missing some values, but on the front end everything is hooked up properly.


ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -1,240 +1,278 @@
...
+  const { id } = useParams();
+
+  const [project, setProject] = useState<ProjectDetails | undefined>(undefined);
+  const [isShowingConfirmDelete, setIsShowingConfirmDelete] = useState(false);
+  const [
+    isShowingConfirmSendReminder,
+    setIsShowingConfirmSendReminder,
+  ] = useState(false);
+  const { error, handleError } = useErrorHandling();
+  const [isLoading, setIsLoading] = useState(true);
```
mtleonard120, line 41: This should likely be initialized as false and then it should be set to true just before the `getProjectById` call below. This way if the id prop did change our component would stay mounted and we would correctly start the loading state again.

PR #114, opened 2020-07-13
CQLHS-85 & CQLHS-237 | Edit existing projects & Client autocomplete

jonathanchaffer:
- `/projects/{id}/edit` fetches the project data for project `id` and autofills the fields.

- [x] Note: I had to manually edit `api.ts` to add `evaluationId` to `EvaluationOverview`. This PR shouldn't be considered done until the swagger docs are updated to actually include this.


jonathanchaffer:
Should be all set. A breaking change was made to the database that will have to be resolved before this can be fully tested

jonathanchaffer:
`story/CQLHS-24` was deleted, so merging this into `develop`

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -109,343 +78,393 @@ export const ProjectEditPage: React.FC = () => {
...
+      values.evaluationDeadline,
+      values.evaluations,
+      values.links,
+      values.notes,
+      values.projectId,
+      values.projectName,
+      values.startDate,
+    ]),
+    [],
+  );
```
charkour, line 147: I'm not confident enough with hooks to know for sure, but can this be achieved without using `useCallback()`, @mtleonard120 ? Also, I would recommend adding cleanup to `useEffect()`, but you can't cleanup when a hook is used within `useEffect()` and I am not sure if `useCallback()` can be cleaned up.

ClientApp/src/components/reuseables/AutoComplete/AutoComplete.tsx
```diff
@@ -26,10 +30,14 @@ export const AutoComplete: React.FC<AutoCompleteProps<object>> = <
   getItemValue,
   getItemId,
   shouldItemRender,
+  defaultValue,
   ...inputProps
 }: AutoCompleteProps<Item>) => {
+  // TODO: should default value be undefined rather than ""?
```
charkour, line 36: The empty string is nice because then `searchText` is typed as a string.

ClientApp/src/components/reuseables/AutoComplete/AutoComplete.tsx
```diff
@@ -1,9 +1,13 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import Autocomplete from "react-autocomplete";
 import Card from "react-bootstrap/Card";
 import "./AutoComplete.scss";
 
+// TODO: use a better autocomplete library
```
charkour, line 6: If so, then a subtask should be added and then the library be added and `AutoComplete` be updated.

ClientApp/src/components/reuseables/AutoComplete/AutoComplete.tsx
```diff
@@ -26,10 +30,14 @@ export const AutoComplete: React.FC<AutoCompleteProps<object>> = <
...
   shouldItemRender,
+  defaultValue,
   ...inputProps
 }: AutoCompleteProps<Item>) => {
+  // TODO: should default value be undefined rather than ""?
   const [searchText, setSearchText] = useState("");
-
+  useEffect(() => {
+    if (defaultValue) setSearchText(defaultValue);
+  }, [defaultValue]);
```
charkour, line 40: Nice to add cleanup here.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
charkour: Currently, isn't `useCallback()` called whenever `values` changes?

charkour:
> `story/CQLHS-24` was deleted, so merging this into `develop`

Yeah, find to target `develop`, CQLHS-24 shouldn't have been deleted.

jonathanchaffer:
> I'm getting a form validation error "All team members must have an employee selected." when I have one PM named Hugh (or any other string). That seems like an issue.

That is expected behavior albeit probably not a great UX. The reason is that you have to select an item from the autocomplete in order for the field to actually have an `Employee` object associated with it. Maybe we could make it so that if you type a valid employee name, it'll search through `employees` for a matching `Employee` and auto-assign it? I currently get a 500 when trying to retrieve Employees from the back end, so something needs to be fixed on their end.

ClientApp/.eslintrc.js
```diff
@@ -101,14 +101,14 @@ module.exports = {
 
     // React Hooks
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
```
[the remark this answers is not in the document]
jonathanchaffer: Like CQLHS-20, I had removed it as a temporary solution then forgot to add it back in. Mike and I discussed this last week, and the rule should definitely stay in. Lots of tutorials online suggest using an empty deps array to mimic `componentDidMount` but it's kind of hack-ish

ClientApp/src/components/pages/LoginPage/LoginPage.tsx
```diff
@@ -33,6 +33,7 @@ export const LoginPage: React.FC = () => {
     errors,
     touched,
   } = useFormik({
+    // TODO: should initial values be undefined rather than ""?
```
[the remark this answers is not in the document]
jonathanchaffer: I think `undefined` makes more sense semantically – an empty string still a string, but if _no string_ has been entered, shouldn't the value be _no string_?

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
mtleonard120: Yeah I would avoid `useCallback`s inside of `useEffect`s as that's not the way you are meant to use them. Rather declare a function using `useCallback` outside of the `useEffect` and then just use that memoized function inside of the `useEffect`, including it in the dependency array.

ClientApp/src/components/reuseables/AutoComplete/AutoComplete.tsx
mtleonard120: In this case a cleanup is not really necessary since there is no async call being made in this effect.

ClientApp/src/components/pages/ProjectEditPage/TeamMemberList/TeamMemberList.tsx
```diff
@@ -1,59 +1,62 @@
...
     );
-    handleUpdate(newTeamMembers);
+    handleUpdate(newEvaluations);
   };
 
-  const [employees, setEmployees] = useState<Employee[]>([]);
+  const [employees, setEmployees] = useState<EmployeeOverview[]>([]);
   const { error, handleError } = useErrorHandling();
 
   useEffect((): void => {
```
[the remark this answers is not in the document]
mtleonard120: This cleanup suggestion is just a little off - you want to do the `isCurrent` check in the `then` and `catch` handlers, since at that point the async call has resolved but you may no longer be current. Make sense?

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -109,343 +78,393 @@ export const ProjectEditPage: React.FC = () => {
...
+                  <Form.Group className="d-flex flex-column" controlId="links">
+                    <Form.Label>
+                      <small>Links</small>
+                    </Form.Label>
+                    <ProjectLinkList
+                      links={values.links || []}
+                      onUpdate={(projectLinks): void =>
+                        setFieldValue("links", projectLinks)
+                      }
+                    />
```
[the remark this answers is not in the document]
jonathanchaffer: The latter is what I was thinking. I don't think it's a bad UX but if you disagree I can change it.

mtleonard120:
@charkour just a heads up, once all the normal dev subtasks on a story are complete and all merged, it's normal process to merge the story branch and delete it. Subsequent bug subtasks added to the story in internal QA can be fixed and have PRs directly targeting develop.

ClientApp/src/components/pages/ProjectEditPage/TeamMemberList/TeamMemberList.tsx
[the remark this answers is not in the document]
charkour: Oh yes, thanks that makes sense.

ClientApp/src/components/reuseables/AutoComplete/AutoComplete.tsx
charkour: Ah alright, thank you!

charkour:
@mtleonard120 , great, thank you! I will keep that in mind. In the case of CQLHS-85, it is a normal story task.

ClientApp/src/components/pages/LoginPage/LoginPage.tsx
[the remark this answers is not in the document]
charkour: Yeah, I'm fine with either and don't know what is _proper_. Go with what makes sense for you!

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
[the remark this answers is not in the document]
charkour: Alright sounds good. Let's leave it and if it feels weird during testing, we can always come back to it.

ClientApp/src/components/pages/ProjectEditPage/TeamMemberList/TeamMemberList.tsx
```diff
@@ -1,83 +1,104 @@
...
   const addTeamMember = (): void => {
-    const newTeamMembers = teamMembers;
-    newTeamMembers.push({
+    const newEvaluations = evaluations;
+    newEvaluations.push({
       employee: undefined,
-      key: nextKey(teamMembers),
+      evaluationId: nextId(evaluations, (evaluation) =>
+        evaluation.evaluationId !== undefined ? evaluation.evaluationId : -1,
+      ),
```
charkour, line 32: Question: When you are making these team members, you're just adding it to the array displayed under "Build the Team" correct? The evaluationId is not sent to the backend and is just used for handling modifications to the array, right?

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -6,101 +6,69 @@ import {
...
     .test(
       "hasManager",
       "Exactly one project manager is required.",
-      (value: TeamMember[]) => {
+      (value: EvaluationOverview[]) => {
         return (
           value.filter(
-            (teamMember) => teamMember.role === ProjectRole.PROJECT_MANAGER,
+            (evaluation) => evaluation.role === ProjectRole.PROJECT_MANAGER,
           ).length === 1
```
charkour, line 45: When editing a project, if the project does not have valid data, `yup` prevents the form from being submitted. Since a user cannot edit the fields producing invalid data, there is an issue. Maybe, `yup` shouldn't check disabled inputs, or maybe we make it so users can edit all fields (rather than having to delete and re-create a project on a mistake in the name or dates or team), or perhaps the UI will fix those invalid fields, the backend will if they receive a project evaluation with invalid fields.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -109,343 +77,388 @@ export const ProjectEditPage: React.FC = () => {
...
+                    <Form.Label>
+                      <small>Additional Notes</small>
+                    </Form.Label>
+                    <textarea
+                      className="form-control"
+                      name="notes"
+                      onChange={handleChange}
+                      placeholder="Notes or thoughts about this project"
+                      rows={3}
+                      value={values.notes || undefined}
```
charkour, line 409: Currently, I am unable to test that editing a project works because I cannot create a project or edit an existing project because of bad data coming from the database.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -109,343 +77,388 @@ export const ProjectEditPage: React.FC = () => {
...
+                    <Col md={4}>
+                      <Row>
+                        <Col>
+                          <Form.Group controlId="client">
+                            <Form.Label>
+                              <small>Client</small>
+                            </Form.Label>
+                            <Form.Control
+                              defaultValue={values.client?.name || undefined}
+                              disabled={!isNewProject}
```
charkour, line 287: Would you be willing to walk through the thought processes of not being able to edit all parts of the project evaluation? That seems like a poor UX and I thought editing all fields were part of the AC? Thanks

ClientApp/src/components/pages/ProjectEditPage/TeamMemberList/TeamMemberList.tsx
jonathanchaffer: Correct, `evaluationId` is also used as the `key` prop for the items in the list

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
jonathanchaffer: I wouldn't mind making all fields editable, but disabling them was a decision made by brandon and ally, so I think we should stick with that. I think the best solution is not to have projects stored on the database that are missing required data.

charkour: Sounds good! Any solution I am fine with.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -109,343 +77,383 @@ export const ProjectEditPage: React.FC = () => {
...
-                            }
-                          />
-                          {touched.reviewDeadline ? (
-                            <Form.Control.Feedback
-                              className="d-block"
-                              type="invalid"
-                            >
-                              {errors.reviewDeadline}
+      <AsyncComponent isLoading={isLoading}>
+        <Form onSubmit={handleSubmit}>
```
charkour, line 150: When clicking "Save Changes" nothing appears to happen. Is `handleSubmit` implemented properly? 
Navigating back to the view screen for the evaluation, the changes do not seem to propagate.

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -109,343 +77,383 @@ export const ProjectEditPage: React.FC = () => {
...
-                      {errors.teamMembers}
-                    </Form.Control.Feedback>
-                  ) : (
-                    <></>
-                  )}
+                  <Form.Group className="d-flex flex-column" controlId="links">
+                    <Form.Label>
+                      <small>Links</small>
+                    </Form.Label>
+                    <ProjectLinkList
```
charkour, line 387: The project link and add link buttons are different heights: 
image

charkour: Maybe this is a bug to be saved for later?

jonathanchaffer: Noted. We can save it for later but I'm guessing it has something to do with the height of the + icon

ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
jonathanchaffer: Yup I noticed that too. Will look into it once the GET /api/Employees fix is merged to develop, since that might have something to do with it

charkour: Sounds good. The endpoint has been fixed and no network request is sent when "Saved Changes" is clicked.

charkour: Also, no request is sent when "Send Evaluation" is clicked when creating a new project eval.

charkour: Ah, I am sorry, I made a mistake during testing. I think "Send Evaluation" works. But will keep testing.

charkour: The backend gives a 500 when creating a new project eval, but a network request is not sent when saving changes for an edit. I think it has something to do with form validation. @jonathanchaffer

jonathanchaffer: Looking into it now

jonathanchaffer: The form was not validating correctly because `notes` was `null`. Setting it to `undefined` fixed it.

PR #115, opened 2020-07-13
CQLHS-76 | Retrieve project & display on Project Evaluations page

jonathanchaffer:
Sample data on the database is currently missing some values, but on the front end everything is hooked up properly.

- [x] Note: I had to manually edit `api.ts` to add a `client` field to `ProjectOverview`. This PR shouldn't be considered done until the swagger docs are updated to actually include this.


ClientApp/.eslintrc.js
```diff
@@ -101,7 +101,6 @@ module.exports = {
 
     // React Hooks
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
```
charkour, line 104: Why is this warning removed?

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -81,27 +93,45 @@ export const useSortContext = (): {
...
-        title="Project Evaluations"
-      />
-      <SortContext.Provider value={{ context, dispatch }}>
-        <ProjectDirectoryTable
-          projects={filterProjectsByTerm(placeholderData, searchText)}
-        />
-      </SortContext.Provider>
+      {isLoading ? (
+        <Spinner animation="border" />
+      ) : (
```
charkour, line 111: This logic is likely going to be repeated on other pages. In the future, some sort of `asyncComponent` component would be helpful to store the loading logic for pages or on-screen widgets.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -81,27 +93,45 @@ export const useSortContext = (): {
...
-        ]}
-        title="Project Evaluations"
-      />
-      <SortContext.Provider value={{ context, dispatch }}>
-        <ProjectDirectoryTable
-          projects={filterProjectsByTerm(placeholderData, searchText)}
-        />
-      </SortContext.Provider>
+      {isLoading ? (
+        <Spinner animation="border" />
```
charkour, line 110: Should the spinner be centered?

ClientApp/.eslintrc.js
charkour: Does this relate to the `useEffect()` on line 100 in `ProjectDirectoryPage.tsx`?

jonathanchaffer: I removed it for a temporary fix and forgot to add it back in. Thanks for pointing that out!

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
jonathanchaffer: Good idea

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
jonathanchaffer: Depends on what Ally says. Wouldn't be a difficult change to make

charkour: Sounds good! Thanks. I'm guessing that would fit best in a future ticket.

Issue #74, opened 2020-07-01
Interacting with a page to add a vertical scroll bar, results in a horizontal one. 


charkour:
This deals with the fact we are using `width: 100vw` to style specific things. Maybe that should only be applied to `html` and then everything else should instead have `100%` for width,

PR #133, opened 2020-07-14
CQLHS-160 | fix styling to prevent content from adding bad scroll bars

charkour:
Currently: 
- if the page is not taller than the view-height, then no vertical scroll bar is added
- if page is taller than view-height, then vertical scroll bar is added
- if interaction with page causes page to become taller than view-height, scroll bar is added and (depending on OS and browser) content will likely shift left a few pixels. 
  - The opposite is true if content becomes shorter than the view-height
- there should never be a horizontal scroll bar 

Question:
- [x] should there always be a vertical scroll bar and never be a horiztional scroll bar?
example code to potentially add:
```
html {
  overflow-y: scroll;
  overflow-x: hidden;
}
```

Addresses: #74


jonathanchaffer:
Lgtm

jonathanchaffer:
Per your question, I don't see any reason why we would want a horizontal scroll bar.

PR #135, opened 2020-07-14
CQLHS-20 | story to develop

jonathanchaffer:


ClientApp/src/styles/index.scss
```diff
@@ -13,12 +13,16 @@ body {
 #root,
 html {
   height: 100vh;
```
charkour, line 15: Not sure how this happened, but line 15 should be `height: 100%` which was changed recently in #133 .
Maybe it is because this got changed on a branch that was also merged into `story/CQLHS-20`?

jonathanchaffer: `height: 100%` caused the login page not to fill the entire page. Changing it to `height: 100vh` fixed that problem and didn't reintroduce the horizontal scroll bar bug as far as I can tell, but let me know if I'm wrong on that.

charkour: Okay, that's great! I forgot to test the login page so thank you for fixing that.

PR #139, opened 2020-07-14
CQLHS-213 | Show the navbar even if a user is not logged in

jonathanchaffer:


ClientApp/.eslintrc.js
```diff
@@ -100,7 +100,7 @@ module.exports = {
     "@typescript-eslint/interface-name-prefix": ["error", "never"],
 
     // React Hooks
-    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/rules-of-hooks": "off",
```
charkour, line 103: Think this change should be reverted, right?

ClientApp/src/components/Navigation/Navigation.tsx
```diff
@@ -26,44 +28,51 @@ export const Navigation: React.FC = () => {
...
+                <>
+                  <Navbar.Toggle aria-controls="navbar-collapse" />
+                  <Navbar.Collapse id="navbar-collapse">
+                    <Nav className="mr-auto">
+                      {user.accessLevel === "admin" ||
+                      user.accessLevel === "teamLead" ? (
+                        <NavLink activeRoute="/dashboard" title="Dashboard" />
+                      ) : (
+                        <></>
+                      )}
```
charkour, line 41: Might consider making these into a component, especially if they get more complex. Don't want to have too much logic in the JSX but it is probably good for now.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
```diff
@@ -1,12 +1,19 @@
...
-export const ProjectDirectoryPage: React.FC = () => (
-  <div className="ProjectDirectoryPage d-flex flex-column justify-content-center align-items-center">
-    <div>
-      <h1 className="display-4">Project Directory</h1>
-    </div>
-    <div>
-      <span>A list of projects at CQL with their project review statuses.</span>
+export const ProjectDirectoryPage: React.FC = () => {
+  requireLogin();
+
```
charkour, line 6: Since this gets added to every page, perhaps we could page our own `Page` component that also includes the `AsyncComponment` as well as part of the boilerplate for each page.

jonathanchaffer: It's not added to every page, just the ones that require authentication. It won't be added to the survey pages or the NotFound page.

ClientApp/.eslintrc.js
jonathanchaffer: It was required to make the `requireLogin` hook in `authService.ts` outside of a functional component.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryPage.tsx
charkour: Ah okay thanks. Perhaps an `AuthPage` is in our future. haha

ClientApp/.eslintrc.js
charkour: Okay, yup I see your thinking with this. However, it is important to [use hooks only within functional components](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions) because it keeps the stateful logic in one place and ensures that the hooks are called in the same order. With the current implementation, I can see the page while I am unauthorized for a split second before it redirects me to log in.

ClientApp/src/services/authService.ts
```diff
@@ -11,3 +14,10 @@ export const validateLogin = async (
   });
   return response.data;
 };
+
+export const requireLogin = (): void => {
+  const history = useHistory();
+  useEffect(() => {
+    if (!loadUser()) history.push("/login");
+  });
+};
```
charkour, line 23: To remove the use of hooks, I would suggest implementing the function similar to this:
```typescript
import { History } from "history";

...

export const requireLogin = (
  history: History<History.PoorMansUnknown>,
): void => {
  if (!loadUser()) history.push("/login");
};
```
and it would be used on the authorized pages similar to `requireLogin(useHistory());`

ClientApp/.eslintrc.js
jonathanchaffer: It's okay to use hooks in [custom hooks](https://reactjs.org/docs/hooks-custom.html). ESLint just didn't know that it was a custom hook. I was able to rename the it to `useRequireLogin` which satisfied the rule

charkour: Ah alright, thanks. That would be a good chance to implement.

PR #148, opened 2020-07-15
project evaluations are now linked to the appropriate responses

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Properties/launchSettings.json
```diff
@@ -15,7 +15,7 @@
       "launchUrl": "api/Projects/",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
-        "CONFIG_CONNECTIONINFO__BENCHMARK": "Server=(local); Database=Benchmark;Trusted_Connection=True;",
```
joeldevries: We should figure out a way for each developer to have their own setting that are not stored in source control so we don't overwrite each others settings. UserSecrets may be an option?

PR #165, opened 2020-07-17
enum bug fix complete

ekm6:


ekm6:
Thanks! Just did it the way you would *shrug emoji here*

PR #168, opened 2020-07-17
employee project evaluations can be filled out

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/SalaryHistory.cs
```diff
@@ -1,7 +1,6 @@
-﻿using Benchmark.WebApi.Models.Database;
-using System;
+﻿using System;
 
-namespace Benchmark.WebApi.Models
+namespace Benchmark.WebApi.Models.Database
 {
     public partial class SalaryHistory
```
tonysulfaro, line 5: are we still utilizing salaryhistory within the app?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Response/Detail/ProjectDetails.cs
```diff
@@ -4,7 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
-namespace Benchmark.WebApi.Models.Response
+namespace Benchmark.WebApi.Models.Response.Detail
```
tonysulfaro, line 7: I noticed it is specifying namespace levels here and other places one level deeper. were additional sub-folders added for the response?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/SalaryHistory.cs
ekm6: nope, I have no idea how it made a comeback

PR #169, opened 2020-07-20
CQLHS-255 | site-wide style updates

jonathanchaffer:
See AC: https://cqlcorp.atlassian.net/browse/CQLHS-255


ClientApp/src/styles/index.scss
```diff
@@ -43,17 +50,37 @@ main {
...
+  .list-group-item {
+    border-color: $table-divider;
+  }
+  .list-group-item.clickable:hover {
+    background-color: $highlight-table-field;
+    cursor: pointer;
+    z-index: 1;
+    border-color: rgba(0, 0, 0, 0);
+  }
+}
```
charkour, line 73: The list items need to be styled so that when they are in the hover state, the divider between elements still shows. Currently, the divider disappears when you hover.

PR #170, opened 2020-07-20
CQLHS-256 | more padding to top of page

jonathanchaffer:


ClientApp/src/styles/index.scss
```diff
@@ -17,7 +17,7 @@ html {
 }
 
 main {
-  padding-top: 2em;
+  padding-top: 50px;
```
charkour, line 20: I know the designs call for static padding and margins, but it's better to make something responsive or reactive. 

For example, this does not look good on mobile. Are we still supporting that?

jonathanchaffer: IIRC, we're supporting mobile but don't need to worry too much about making everything look good on mobile.

charkour: Cool thanks.

PR #173, opened 2020-07-20
CQLHS-260 & CQLHS-261 | project editor & link modal updates

jonathanchaffer:


ClientApp/src/components/pages/ProjectEditPage/ProjectLinkModal/ProjectLinkModal.tsx
```diff
@@ -90,25 +104,28 @@ export const ProjectLinkModal: React.FC<ProjectLinkModalProps> = ({
...
+              onClick={name ? handleReset : onDelete}
               variant="outline-info"
             >
               Cancel
             </Button>
-            <Button className="mr-2" onClick={onDelete} variant="danger">
-              Delete
-            </Button>
-            <Button type="submit">Save</Button>
+            <Button type="submit">{name ? "Update Link" : "Add Link"}</Button>
```
charkour, line 128: I don't think these changes broke adding the links. But there is a bug where you can only add one new link and if you try to add a second new link, the first one disappears. 

Either, it needs to be addressed in this PR or another bug ticket should be made.

jonathanchaffer: Hmm, I'm not getting that on my machine. Steps to reproduce?

PR #175, opened 2020-07-20
client questions now added to all evaluations

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/Project.cs
```diff
@@ -19,7 +19,7 @@ public Project(ProjectUpdateRequest request)
             ClientId = request.ClientId;
             Deadline = request.EvaluationDeadline;
             EndDate = request.EndDate;
-            StartDate = request.EndDate;
+            StartDate = request.StartDate;
```
tonysulfaro, line 22: good catch

PR #176, opened 2020-07-20
CQLHS-186 | display employee's project eval results

charkour:
Displays them with notes and an accordion dropdown.

- Styling will come in future subtask
- Calculating score will be done on the backend
- Selecting one employee from a drop down will be in a future subtask

Question: How to style the `ListGroup` with the `variant="flush"` when passed as a custom prop for `Accordion`?

Update: 
- Add ability to customize title and subtitle on eval table
- Fix not being able to close all parts of accordion menu.


ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -28,18 +25,24 @@ export const ProjectDetailPage: React.FC = () => {
...
   const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
     setIsLoading(true);
     getProjectById(parseInt(id))
-      .then((proj) => setProject(proj))
+      .then((proj) => {
+        setProject(proj);
+        getProjectEvaluationResults(proj?.projectId)
+          // TODO: Should this use async/await?
```
jonathanchaffer, line 42: Re: this TODO I think they're two ways of doing the same thing. I tend to prefer `.then().catch().finally()` but that's mostly cause I'm more used to it.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationResultsTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <Col>
+            <h6>Question</h6>
+          </Col>
+          <Col>
+            <h6>Average Response</h6>
+          </Col>
+        </Row>
+      </Card.Header>
+      {/* TODO: how ot give ListGroup the variant="flush" style when given as "as" */}
+      <Accordion as={ListGroup}>
```
jonathanchaffer, line 33: You can add `className="list-group-flush"` to the `Accordion` to achieve this

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -0,0 +1,50 @@
...
+                </Col>
+              </Row>
+              <Row>
+                <Col>
+                  <EvaluationResultsTable section={section} />
+                </Col>
+              </Row>
+            </Col>
+            <Col md={4} xs={12}>
+              {/* TODO: how to do spacing? */}
```
jonathanchaffer, line 38: I would suggest putting the `EvaluationNotes` in the same `Row` as the `EvaluationResultsTable`. Then they'll align to each other.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -0,0 +1,50 @@
...
+              <Row>
+                <Col>
+                  <h3>Evaluation Results</h3>
+                </Col>
+              </Row>
+              <Row>
+                <Col>
+                  <h4>{`Results for ${section.employeeReviewee.firstName} ${section.employeeReviewee.lastName}`}</h4>
+                </Col>
+              </Row>
```
jonathanchaffer, line 30: Only one heading should be here, and a subheading for the employee's project role.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
jonathanchaffer: Something like: 
```
    <>
      {results?.sections.map((section) => (
        <>
          <Row>
            <Col>
              <h4>{`Results for ${section.employeeReviewee.firstName} ${section.employeeReviewee.lastName}`}</h4>
            </Col>
          </Row>
          <Row>
            <Col md={8} xs={12}>
              <EvaluationResultsTable section={section} />
            </Col>
            <Col md={4} xs={12}>
              <EvaluationNotes notes={section.notes} />
            </Col>
          </Row>
        </>
      ))}
    </>
```

ClientApp/src/components/pages/ProjectDetailPage/ProjectDetailPage.tsx
charkour: Yup, that makes sense. Hopefully, we won't have to nest too many requests.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationResultsTable.tsx
charkour: Thanks

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
charkour: Yeah, that's for the employee view. I'll make my component reusable so that it can be used in both scenarios.

PR #177, opened 2020-07-20
CQLHS-258 | project directory updates

### LINKED ISSUE #158: Client column should sort A-Z by default on the Project Directory page
Currently it sorts by Z-A. The values being compared in the sorting function need to be swapped.

jonathanchaffer:
See AC: https://cqlcorp.atlassian.net/browse/CQLHS-258

Fixes #158


ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
```diff
@@ -69,33 +87,35 @@ export const ProjectDirectoryTable: React.FC<ProjectDirectoryTableProps> = ({
...
+        </Col>
+        <Col lg={1} sm={2} xs={3}>
+          <ProjectTableSortButton attribute={ProjectSortable.START_DATE} />
+        </Col>
+        <Col lg={1} sm={2} xs={3}>
+          <ProjectTableSortButton attribute={ProjectSortable.END_DATE} />
+        </Col>
+        {/* TODO: only render the following element if user is admin */}
+        <Col className="d-none d-lg-block" xs={2}>
+          <ProjectTableSortButton attribute={ProjectSortable.STATUS} />
```
charkour, line 115: Some weird sorting happens when you try to sort by status. It will sort by end date (I think?) every two clicks. 
Are we supposed to have data for status yet?

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
```diff
@@ -24,20 +26,36 @@ interface ProjectDirectoryTableProps {
...
+            >
+              <CloseIcon fontSize="small" />
+              Reset
+            </Button>
+          </Col>
+        </Row>
+      ) : (
+        <></>
+      )}
+
```
charkour, line 50: Currently, making the Reset button hide and show results in the table jumping up and down. I think it'll behave better if the space is constant and does not jump.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
jonathanchaffer: I think it's sorting them either in the order they're stored in the array, or arbitrarily, since there all of the statuses are the same so it doesn't know how to sort them. Should be fixed once the backend is sending back actual data for that field.

ClientApp/src/components/pages/ProjectDirectoryPage/ProjectDirectoryTable.tsx
jonathanchaffer: Added animation to make it less jerky.

PR #189, opened 2020-07-21
CQLHS-209 | retrieve & render EvaluationDetails from GUID

jonathanchaffer:


ClientApp/.eslintrc.js
```diff
@@ -80,7 +80,6 @@ module.exports = {
     "implicit-arrow-linebreak": "off",
 
     // Alphabetize
-    "sort-keys": "error",
```
jonathanchaffer, line 83: Removed this because it conflicts with `sort-keys-fix` – one puts capital letters first and the other puts lowercase letters first.

ClientApp/src/components/pages/EvaluationSurveyPage/EvaluationSurvey.tsx
```diff
@@ -0,0 +1,275 @@
...
+                      dispatch({
+                        payload: { index: currentIndex + 1 },
+                        type: "setIndex",
+                      });
+                    }}
+                  >
+                    Next
+                  </Button>
+                )}
+              </>,
```
charkour, line 236: There is a lot going on here to create the elements in the `actions` array. I think it would be best suited to abstract this code to be used somewhere else and just pass identifiers to the array.

ClientApp/src/components/pages/EvaluationSurveyPage/EvaluationSectionView/EvaluationSectionView.tsx
```diff
@@ -64,43 +62,41 @@ export const EvaluationSection: React.FC<EvaluationSectionProps> = ({
...
+                    <Col className="d-flex align-items-center" md={8} xs={12}>
+                      <Row className="flex-fill">
+                        {responses.map((response, index) => (
+                          <Col
+                            key={response}
+                            className="d-flex justify-content-center"
+                            xs={2}
+                          >
+                            <Form.Check
+                              id={uuid()}
```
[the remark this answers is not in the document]
jonathanchaffer: Every radio button in the survey needs a unique id so clicking one doesn't erroneously select other radio buttons, which is why I used `uuid()`, but I suppose that's a bit overkill. Updated so the id is a combination of the target employee id, question id, and response.

PR #195, opened 2020-07-21
CQLHS-189 | Select employee project eval results dropdown

charkour:
- updates `react-bootstrap` to 1.2.2 to use context related features for the accordion. See [discussion here](https://github.com/react-bootstrap/react-bootstrap/issues/4140).
  - didn't end up needing to use `AccordionContext` but trying to downgrade version did not revert the bugs caused by the upgrade
  - causes type errors with `formik` and `react-bootstrap` that are now fixed
- create a custom component and hook to determine if an accordion toggle is open
- add ability to select an employee project evaluation results from drop down

- styling will come in future subtask
- calculating the average, done by backend.


ClientApp/src/components/pages/LoginPage/LoginPage.tsx
```diff
@@ -81,7 +81,11 @@ export const LoginPage: React.FC = () => {
           <div className="divider h" />
         </Col>
         <Col md={5}>
-          <Form onSubmit={handleSubmit}>
+          <Form
+            onSubmit={(event: FormEvent<HTMLElement>): void =>
+              handleSubmit(event as FormEvent<HTMLFormElement>)
+            }
+          >
```
jonathanchaffer, line 88: I suspect that a future Formik update will fix the type inconsistencies, but for now, would it be possible to move this logic to a `handleSubmit` function, since it's used several places throughout the app?

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -17,33 +19,86 @@ export const ProjectEvaluationResults: React.FC<ProjectEvaluationResultsProps> =
...
+
+const EvaluationDropdown: React.FC<EvaluationDropdownProps> = ({
+  results,
+  title,
+  onItemClick,
+}: EvaluationDropdownProps) => {
+  return (
+    <>
+      Results for
+      <DropdownButton id="dropdown-basic-button" title={title} variant="info">
```
jonathanchaffer, line 86:
```suggestion
      <DropdownButton id="dropdown-basic-button" title={title} variant="outline-info">
```

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -17,33 +19,86 @@ export const ProjectEvaluationResults: React.FC<ProjectEvaluationResultsProps> =
...
+        {results?.sections.map(
+          ({ employeeReviewee: { employeeId: id, firstName, lastName } }) => (
+            <Dropdown.Item
+              key={id}
+              eventKey={`${id}`}
+              onClick={(): void => {
+                // TODO: pretty sure this will change to fullName on the api.
+                onItemClick(`${firstName || ""} ${lastName || ""} `, id ?? -1);
```
jonathanchaffer, line 94: For now, you can use `fullName` from `utilities/stringTransform.ts`.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -17,33 +19,86 @@ export const ProjectEvaluationResults: React.FC<ProjectEvaluationResultsProps> =
...
+              }}
+            >
+              {firstName} {lastName}
+            </Dropdown.Item>
+          ),
+        )}
+      </DropdownButton>
```
jonathanchaffer, line 101: `react-bootstrap` has multiple ways of making a dropdown. To be consistent with other dropdowns in the app, I would suggest 
```suggestion
      <Dropdown>
        <Dropdown.Toggle id="dropdown-basic-button" variant="info">
          {title}
        </Dropdown.Toggle>
        <Dropdown.Menu>
          {results?.sections.map(
            ({ employeeReviewee: { employeeId: id, firstName, lastName } }) => (
              ...
            ),
          )}
        </Dropdown.Menu>
      </Dropdown>
```

ClientApp/src/components/reuseables/Tooltipped/Tooltipped.tsx
```diff
@@ -3,7 +3,7 @@ import OverlayTrigger from "react-bootstrap/OverlayTrigger";
 import Tooltip from "react-bootstrap/Tooltip";
 
 interface TooltippedProps {
-  children: React.ReactNode;
+  children: JSX.Element;
```
jonathanchaffer, line 6: I'm not sure what the difference is between these. Is `JSX.Element` better? If so we'll probably want to change it in `HeaderProps` too.

ClientApp/src/placeholders/ProjectEvaluations.ts
```diff
@@ -216,12 +216,12 @@ export const placeholderEvaluationResults: ProjectEvaluationResultsData = {
...
         {
           employeeReviewer: placeholderEmployeeOverviews[1],
-          note: "a note also",
+          note: undefined,
           noteId: 22,
         },
         {
           employeeReviewer: placeholderEmployeeOverviews[2],
-          note: "a third note",
+          note: "a third note wow so cool love you so much",
```
jonathanchaffer, line 224: lol

ClientApp/src/components/reuseables/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./AutoComplete";
 export * from "./ConfirmModal";
+export * from "./ContextAwareToggle";
```
jonathanchaffer, line 3: Maybe rename `ContextAwareToggle` to `ContextAwareAccordionToggle`?

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -17,33 +19,86 @@ export const ProjectEvaluationResults: React.FC<ProjectEvaluationResultsProps> =
...
+          ({ employeeReviewee: { employeeId: id, firstName, lastName } }) => (
+            <Dropdown.Item
+              key={id}
+              eventKey={`${id}`}
+              onClick={(): void => {
+                // TODO: pretty sure this will change to fullName on the api.
+                onItemClick(`${firstName || ""} ${lastName || ""} `, id ?? -1);
+              }}
+            >
+              {firstName} {lastName}
```
jonathanchaffer, line 97: For now, you can use `fullName` from `utilities/stringTransform.ts`.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -17,33 +19,86 @@ export const ProjectEvaluationResults: React.FC<ProjectEvaluationResultsProps> =
...
+  results: ProjectEvaluationResultsData | undefined;
+  title: string;
+}
+
+const EvaluationDropdown: React.FC<EvaluationDropdownProps> = ({
+  results,
+  title,
+  onItemClick,
+}: EvaluationDropdownProps) => {
+  return (
```
jonathanchaffer, line 83: Maybe add
```suggestion
}: EvaluationDropdownProps) => {
const sections = results?.sections;
  return (
```
It's only used once so I'm not sure how critical it is, but wouldn't hurt for readability.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.scss
```diff
@@ -1,3 +1,5 @@
 .accordion > .list-group-item {
   cursor: pointer;
+  // TODO: remove text highlight
+  // ref: https://stackoverflow.com/questions/826782/how-to-disable-text-selection-highlighting
```
jonathanchaffer, line 4: Is there a reason you want to disable highlighting?

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
charkour: Ah, that's where it is. I was looking for it. Thanks

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.scss
charkour: It was just annoying when double-clicking the toggles to have it highlight. I'll take it off for now and see what Ally says in QA.

ClientApp/src/components/reuseables/Tooltipped/Tooltipped.tsx
charkour: `JSX.Element` is stricter than `React.ReactNode` in it's typing. FCs return `JSX.Element` so I believe that is why it can only take those in as children. I don't totally understand the nuances, but here is a [discussion that goes into some of the details](https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement).

ClientApp/src/utilities/stringTransform.ts
```diff
@@ -3,8 +3,8 @@ export const standardize = (str: string): string => {
 };
 
 export const fullName = (
-  first: string | undefined,
-  last: string | undefined,
+  first: string | null | undefined,
+  last: string | null | undefined,
```
jonathanchaffer, line 7: To avoid nulls, I've just been invoking `fullName` like `fullName(firstName || undefined, lastName || undefined)`. I'm fine with the change you made, but you may want to update all existing invocations of `fullName` to not use `|| undefined`

charkour: Ah, alright thanks.

PR #204, opened 2020-07-22
CQLHS-284 & CQLHS-287 | email bg color; header height

jonathanchaffer:


Benchmark.WebApi/Benchmark.WebApi/Emails/email-template.html
```diff
@@ -77,13 +77,13 @@
 </head>
 <!-- Edit the code below this line -->
 
-<body class="bg-light">
-  <div class="pt-4 container" style="height: 100vh;">
+<body class="bg" style="background-color: #f5f5f5;">
+  <div class="pt-4 container" style="height: 100vh">
```
charkour, line 81: Don't think missing this semicolon will break the CSS

PR #211, opened 2020-07-23
CQLHS-292 | fix project save bug and add explicit types

charkour:
Also adds another workspace to the vscode settings.

Note: large changes are due to changing indentation. 

This requires that you remove and re-install `node-modules/ `


charkour:
After the current, front-end PRs are merged, this will need to be updated.

PR #212, opened 2020-07-23
CQLHS-144 | skeleton UI for employee profile page

jonathanchaffer:


jonathanchaffer:
Fixed it. thanks for the tip!

PR #215, opened 2020-07-23
CQLHS-329 | remove link id from POST /api/Projects/save request

jonathanchaffer:


ClientApp/src/services/projectService.ts
```diff
@@ -189,9 +189,13 @@ export const sendProjectEvaluation = async (
...
-    links: project.links?.filter(
-      (link: LinkSnippet): boolean => link.url !== undefined,
-    ),
+    links: project.links
+      ?.filter((link: LinkSnippet): boolean => link.url !== undefined)
+      .map(
+        (link: LinkSnippet): LinkSnippet => {
+          return { ...link, linkId: undefined };
+        },
+      ),
```
charkour, line 198: A linkId needs to be returned when updating a link. Currently, edited links are resulting in completely new links being created. A linkId should not be returned when it is a completely new link. 

To reproduce. Create project will link. Go to project detail page and look at the linkId in the response. Edit the project and then save it. Go back to the project detail page and look at the linkId again and it will be different than before.

jonathanchaffer: Ah, good catch. I forgot that creating and updating projects were the same request.

ClientApp/src/services/projectService.ts
```diff
@@ -169,12 +169,16 @@ export const deleteProject = (): void => {
...
+  isNew: boolean,
 ): Promise<void> => {
   const manager =
     project.evaluations?.find(
       (evaluation: EvaluationOverview): boolean =>
         evaluation.projectRole === ProjectRole.PM,
     )?.employee || undefined;
+  const chosenLinks = project.links?.filter(
+    (link: LinkSnippet): boolean => link.url !== undefined,
+  );
```
charkour, line 181: Good, but not quite. 

If you add a link to an existing project, then the server will return a 500. 
This needs to check if the linkId is undefined (an added link) or not undefined (an updated link). This might require changing the `ProjectLink` component since I think you said project links have ids that are different than the database ids? You may want to separate those two so they don't cause logical bugs.

PR #238, opened 2020-07-27
CQLHS-188 | map backend data to project eval results table

charkour:
To test, go to `projects/147` and the survey data will be from the demo last Friday. 
The data received is different than expected, so that is why there are sometimes missing people in notes or the notes say NA rather than - . 
If those are issues, then a bug ticket will need to be submitted for the backend.


ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -101,11 +94,10 @@ const EvaluationDropdown: React.FC<EvaluationDropdownProps> = ({
...
             ({
-              employeeReviewee: { employeeId: id, firstName, lastName },
-            }: ProjectEvaluationEmployeeResult): JSX.Element => {
-              const name = fullName(firstName, lastName);
-              // TODO: update the project role when backend model supports it
-              const itemTitle = `${name} - Project Role`;
+              targetEmployeeSnippet: { employeeId: id, fullName } = {},
+              targetProjectRole: role,
+            }: EvaluationResultSection): JSX.Element => {
+              const itemTitle = `${fullName} - ${role}`;
```
jonathanchaffer, line 100: To display roles properly, this should either be
```suggestion
              const itemTitle = `${fullName} - ${transformProjectRole(role as ProjectRole)}`;
```
or...
```suggestion
              const itemTitle = `${fullName} - ${transformProjectRole(role)}`;
```
For the latter, the api would need to be updated to return a `ProjectRole` rather than a string.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
```diff
@@ -1,18 +1,17 @@
-import { ProjectEvaluationEmployeeNote } from "placeholders";
 import React from "react";
 import Card from "react-bootstrap/Card";
 import Col from "react-bootstrap/Col";
 import ListGroup from "react-bootstrap/ListGroup";
 import Row from "react-bootstrap/Row";
-import { fullName } from "utilities";
+import { EvaluationNotes as EvaluationNote } from "_generated";
```
jonathanchaffer, line 6: Although I do think `EvaluationNote` is a better name than `EvaluationNotes`, renaming imports is a bit of an anti-pattern IMO. If it's important for it to be named `EvaluationNote`, we should ask the backend team to update the api.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
```diff
@@ -1,18 +1,17 @@
...
 
-interface EvaluationNotesProps {
-  notes: ProjectEvaluationEmployeeNote[];
+interface EvaluationNotesTableProps {
+  notes: EvaluationNote[] | undefined | null;
```
jonathanchaffer, line 9: To deal with nulls as little as possible, I would recommend
```suggestion
  notes: EvaluationNote[] | undefined;
```
Then when passing props, use `notes || undefined` to cast any falsy valye to `undefined`.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
```diff
@@ -23,20 +22,20 @@ export const EvaluationNotes: React.FC<EvaluationNotesProps> = ({
...
+        {notes?.map(
           ({
-            noteId: id,
-            note,
-            employeeReviewer: { firstName, lastName },
-          }: ProjectEvaluationEmployeeNote): JSX.Element => (
-            <ListGroup.Item key={id}>
+            content,
+            notesId,
+            employee: { fullName } = {},
```
jonathanchaffer, line 29: Object destructing is good for making code more readable, but when you start to destruct optional attributes the syntax gets a bit confusing. Personally I'd prefer
```suggestion
            employee,
```
and update line 34 to just use `employee?.fullName`.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
```diff
@@ -23,20 +22,20 @@ export const EvaluationNotes: React.FC<EvaluationNotesProps> = ({
...
+            <ListGroup.Item key={notesId}>
               <Row className="d-flex align-items-center">
                 <Col>
-                  <h6>{fullName(firstName, lastName)}</h6>
+                  <h6>{fullName}</h6>
                 </Col>
               </Row>
               <Row className="d-flex align-items-center">
-                <Col>{note || "-"}</Col>
+                <Col>{content || "-"}</Col>
```
jonathanchaffer, line 38: If back end team isn't able to update what's returned for no notes, then this should be updated to:
```suggestion
                <Col>{content !== "NA" ? content || "-" : "-"}</Col>
```
But it would be better for the back end to change since doing it this way would ignore an edge case where a user actually types "NA" into the notes field.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationResultsTable.tsx
```diff
@@ -43,44 +43,47 @@ export const EvaluationResultsTable: React.FC<EvaluationResultsTableProps> = ({
...
+          },
         )}
       </Accordion>
     </Card>
   );
 };
 
 interface EvaluationSubTableProps {
-  responses: ProjectEvaluationQuestionResponse[];
+  scores: QuestionResponseSnippet[] | null | undefined;
```
jonathanchaffer, line 78: See my previous comment about eliminating nulls.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
```diff
@@ -77,16 +71,15 @@ export const ProjectEvaluationResults: React.FC<ProjectEvaluationResultsProps> =
 
 interface EvaluationDropdownProps {
   onItemClick: (name: string, id: number) => void;
-  results: ProjectEvaluationResultsData | undefined;
+  sections: EvaluationResultSection[] | undefined | null;
```
jonathanchaffer, line 74: See my previous comment about eliminating nulls.

ClientApp/src/components/reuseables/ProjectEvaluationResults/ProjectEvaluationResults.tsx
charkour: Thanks!

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
charkour: Yeah, I meant to remove that. Thanks.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
charkour: Good call

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
charkour: Backend will change it.

ClientApp/src/components/reuseables/ProjectEvaluationResults/EvaluationNotesTable.tsx
charkour: Sure thing. I kinda like the one-liners :P

PR #247, opened 2020-07-27
Story/cqlhs 267

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/GoalsController.cs
```diff
@@ -11,7 +12,7 @@
 namespace Benchmark.WebApi.Controllers
 {
     [Route("api/[controller]")]
-    [Authorize]
+    [AllowAnonymous]
```
tonysulfaro, line 15: Can you remove this before going into dev?

PR #251, opened 2020-07-28
CQLHS-351, CQLHS-354, CQLHS-363 | automatically reauthenticate users

jonathanchaffer:
- New access token is auto-generated every 5 minutes
- User is reauthenticated immediately upon first render to prevent users from changing their role by editing local storage
- If a 401 is encountered, user is logged out and returned to login screen. This will rarely ever happen thanks to the automatic reauthentication every 5 minutes


jonathanchaffer:
Originally I had it so when a 401 was received, the app would request a new access token. However, I was getting a 404 from the request because the access token was no longer valid – the endpoint requires a valid refresh token AND a valid access token, which I think is a flaw in the API design, since it kind of defeats the purpose of having a refresh token at all. I met with the back end team and they suggested re-authenticating every 5 minutes, which is a working solution, albeit unconventional. So I guess it's a matter of whether we want to go thru the effort of updating both the back end and front end, or keeping it as-is.

charkour:
Ah okay, thanks for the explanation. I agree the current API is unconventional, so we will keep it as is.

PR #252, opened 2020-07-28
super user access to changing role

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/AuthService.cs
```diff
@@ -23,6 +23,7 @@ public class AuthService : IAuthService
         private readonly IEmployeeRepository _employeeRepository;
         private readonly IActiveDirectoryRepository _activeDirectoryRepository;
         private readonly JWTSettings _jwtsettings;
+        private readonly ICollection<string> _superUsers = new List<string>() { "john.chaffer", "charlie.kornoelje", "ally.brooks", "enoch.mwesigwa", "tony.sulfaro" };
```
joeldevries, line 26: Hardcoded usernames are not a good idea. I would consider 1 of two other options:
1. Make "Super User" a role that users can have - that would then allow them to switch to another role. Or,
2. Put this list of users in the config file so we can change it in different environments without recompiling.

ekm6: ah, that makes sense. Thanks!

PR #257, opened 2020-07-28
CQLHS-307 | Fetch & display the list of employees, separated by TL

charkour:
The data is fetched and displayed on a table. 
There are a few todos, one waiting for the backend, and one for styles which will come in a later subtask. 

Is there a better way to map through the groups than I used?


ClientApp/src/styles/index.scss
```diff
@@ -255,3 +255,11 @@ $dot-size: 1.25em;
     position: relative;
   }
 }
+
+.TableBody span {
```
jonathanchaffer, line 259: `TableBody` is not a great class to apply this style to – there may be cases where we want a table body that doesn't have the overflow ellipsis behavior. My suggestion would be
```suggestion
.overflow-ellipsis,
.overflow-ellipsis span {
```
and apply the class `overflow-ellipsis`, rather than `TableBody` to the div. Also, CSS class names usually use `kebab-case` – the only reason we have some in `PascalCase` is to match the names of React components.

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+                <Col>
+                  <span>{moment(hireDate).format("MM/DD/YYYY")}</span>
+                </Col>
+                <Col>
+                  {/* TODO: Need a way to see if the review is in progress */}
+                  <span>{moment(lastReviewDate).format("MM/DD/YYYY")}</span>
+                </Col>
+                <Col>
+                  <Button>View Progress</Button>
+                </Col>
```
jonathanchaffer, line 83: This column can be removed from the table since it does the same thing as clicking the `list-group-item` itself, and may cause confusion. Ally approved this change in https://projects.invisionapp.com/d/main?origin=v7#/console/20116666/422170801/comments/134289771

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
```diff
@@ -1,17 +1,91 @@
...
+              <Col>
+                <h3>Teams</h3>
+              </Col>
+            </Row>
+            {Object.entries(sortAndGroup(employees) || []).map(
+              (group: [string, EmployeeOverview[]]): JSX.Element => {
+                const managerName = group[0];
+                const groupedEmps = group[1];
+                return (
+                  <React.Fragment key={managerName + groupedEmps}>
```
jonathanchaffer, line 70: TIL that `<></>` is a shorthand for `React.Fragment`!

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
```diff
@@ -1,17 +1,91 @@
...
-import { useRequireLogin } from "services";
-import { PermissionType } from "_generated";
+import {
+  AsyncComponent,
+  ErrorModal,
+  Header,
+  PermissionsWrapper,
+  useErrorHandling,
+} from "components";
+import _ from "lodash";
```
jonathanchaffer, line 8: Perhaps this identifier could be more descriptive?
```suggestion
import lodash from "lodash";
```

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
```diff
@@ -1,17 +1,91 @@
...
+  const [employees, setEmployees] = useState<EmployeeOverview[]>([]);
+  const { error, handleError } = useErrorHandling();
   useRequireLogin();
+
+  const sortAndGroup = (
+    emps: EmployeeOverview[],
+  ): _.Dictionary<EmployeeOverview[]> => {
+    return _(emps)
+      .sortBy(
+        (e: EmployeeOverview): string | null | undefined => e.manager?.fullName,
```
jonathanchaffer, line 29: To avoid nulls,
```suggestion
        (e: EmployeeOverview): string | undefined => e.manager?.fullName || undefined,
```
Similarly on line 32.

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
```diff
@@ -1,17 +1,91 @@
...
+      )
+      .groupBy(
+        (e: EmployeeOverview): string | null | undefined => e.manager?.fullName,
+      )
+      .value();
+  };
```
jonathanchaffer, line 35: I like the use of lodash to sort and group employees. I was hoping the grouping would be done on the back end, but as long as this is efficient I think it's fine. We may also want to update the `groupProjectsByYear` function used on the Project Directory page – currently it's just a function I implemented manually, which might not be as efficient as this method.

ClientApp/src/styles/index.scss
charkour: Good call-out. Thanks!

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
charkour: haha the _ is a "low dash". But yeah, I can use better imports that don't import everything.

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
charkour: Yeah, I'm not sure what is more efficient between lodash or your localeCompare.

PR #265, opened 2020-07-29
Remove WebDAV module and handler which is blocking PUT and DELETE req…

joeldevries:
…uests.


joeldevries:
LOL - looked fine in my Git client:

image

PR #267, opened 2020-07-29
CQLHS-379 | add cleanup to async useEffect

### LINKED ISSUE #227: Add cleanup to useEffect when doing an async request
Some useEffect functions call async functions that are not properly cleaned up and will try to set state to an unmounted component.

charkour:
No mem leaks from async calls nows.

There has to be a better way to manage this.

Fixes #227


ClientApp/src/components/pages/ProjectEditPage/ProjectEditPage.tsx
```diff
@@ -135,24 +135,35 @@ export const ProjectEditPage: React.FC = (): JSX.Element => {
```
jonathanchaffer, line 135: Can you update line 115 to initialize `isLoading` to false, and then set it to true at the beginning of `useEffect`?

charkour: Not sure exactly what it is referring to. But it is fine to keep the loading state in the callback since it is the same structure as all of the requests I would say.

PR #288, opened 2020-07-31
CQLHS-402 | add goal description

tonysulfaro:
migration works. has to be run on dev before merge though


tonysulfaro:
This became out of date with pr #289

PR #291, opened 2020-07-31
CQLHS-318: display details of a specific goal

jonathanchaffer:
Draft until TODOs are resolved


jonathanchaffer:
One remaining TODO: `GoalUpdateSnippet` needs a `title` field. Requesting review before that's resolved so it doesn't block other subtasks

jonathanchaffer:
Also, the Goal Information doesn't totally match design comps, but it was approved by Ally.

PR #296, opened 2020-07-31
CQLHS-309 | Add the ability to edit team members for a TL

charkour:
Styling will come in new subtask.
Assigning an unassigned employee to a non-existing team will come in a new subtask

This gives the ability to update team members.
Updated the ConfirmationModal to take jsx in it's body. 

- [x] Also, I have no idea why CQLHS-307 commits are showing up here because they got merged into the story here #257 
- it is due to rebasing. The changes from the commits won't show up in the diff which is good.


ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+        isShowing={isConfirmModalShowing}
+        onCancel={onConfirmationCancel}
+        onConfirm={onConfirmationConfirm}
+      />
+      <ErrorModal error={error} />
+    </div>
+  );
+};
+
+interface EditTeamModalBody {
```
jonathanchaffer, line 170:
```suggestion
interface EditTeamModalBodyProps {
```

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
```diff
@@ -1,43 +1,77 @@
...
+  } = useModal<EmployeeOverview[]>();
   useRequireLogin();
 
+  /**
+   * Used when sorting and grouping employees.
+   */
+  const getManagerName = (e: EmployeeOverview): string => {
+    if (e.manager?.employeeId === e.employeeId) return "Unassigned";
+    return e.manager?.fullName || "";
+  };
```
jonathanchaffer, line 37: Semantically I think this would make more sense:
```suggestion
  const getManagerName = (e: EmployeeOverview): string | undefined => {
    if (e.manager?.employeeId === e.employeeId) return undefined;
    return e.manager?.fullName || undefined;
  };
```
Not that it would ever happen, but this would fix an edge case where someone's actual name is "Unassigned". More importantly, you wouldn't have to hardcode the string "Unassigned" in other places where you're checking for an employee's manager.

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
```diff
@@ -1,43 +1,77 @@
...
+   */
+  const getManager = (
+    emps: EmployeeOverview[] | undefined,
+  ): EmployeeOverview => {
+    return (
+      (!!emps &&
+        emps[0].employeeId !== emps[0].manager?.employeeId &&
+        emps[0].manager) || { employeeId: -1, fullName: "Unassigned" }
+    );
+  };
```
jonathanchaffer, line 74: Similar to my previous comment, if there is no manager, it makes more sense to return undefined than a "fake" manager.
```suggestion
  const getManager = (
    emps: EmployeeOverview[] | undefined,
  ): EmployeeOverview | undefined => {
    return (
      (!!emps &&
        emps[0].employeeId !== emps[0].manager?.employeeId &&
        emps[0].manager) || undefined
    );
  };
```

ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <>
+      <>Are you sure you want to submit these team assignments?</>
+      {reassigned.length > 0 && (
+        <>
+          These employees will be reassigned:
+          {reassigned.map(
+            ({ employeeId: id, fullName }: EmployeeOverview): JSX.Element => (
+              <div key={id}>
+                {`Name: ${fullName}; Current Manager: ${manager?.fullName}`}
+              </div>
```
jonathanchaffer: This currently shows the newly assigned manager, but I think it should show the old manager.

ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+    postEmployeesAssign(team, assigned, managerId)
+      .then((): void => {
+        setIsConfirmModalShowing(false);
+        if (onRefresh) onRefresh();
+      })
+      .catch((err: Error): void => handleError(err))
+      .finally((): void => setIsLoading(false));
+  };
+
+  const confirmModalBody = (
```
jonathanchaffer, line 102: Might also want to show the team members that have been unassigned from the manager in this confirmation modal.

ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+            ),
+          )}
+        </>
+      )}
+    </>
+  );
+
+  return (
+    // TODO: update this modal (or create a new one), that allows
+    //  unassigned employees to be assigned to a team that does not exist yet.
```
jonathanchaffer, line 124: Re: this todo: I think the modal for unassigned employees is distinct enough from the existing one to warrant creating a new modal.

ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
charkour: Thanks. Didn't realize that you could name a function and an interface the same thing.

ClientApp/src/components/pages/EmployeeDirectoryPage/EmployeeDirectoryPage.tsx
charkour: Thanks for the comment. With this, I would push back because returning undefined here will just cause more inline ternaries which will make the JSX more complex than needed. I prefer to get rid of potential `null | undefined`  through abstractions. In this part of the logic, all employees who are unassigned are effectively under the same unassigned manager.

ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
charkour: Good catch. Missed that bug.

ClientApp/src/components/reuseables/EditTeamModal/EditTeamModal.tsx
charkour: Good call.

PR #309, opened 2020-08-03
query complete with exception of annual reviews (yet to be solidified)

ekm6:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Controllers/EmployeesController.cs
```diff
@@ -15,17 +15,25 @@
 namespace Benchmark.WebApi.Controllers
 {
     [Route("api/[controller]")]
-    [Authorize]
+    [AllowAnonymous]
```
tonysulfaro, line 18: just make sure to remove this before it goes into develop
```suggestion
    [Authorize]
```

PR #320, opened 2020-08-04
CQLHS-412 | add unassigned employee to team

charkour:
~Currently a work in progress. Blocked by backend endpoints.~

To test, login and go to the employee directory and click "assign employees".


ClientApp/src/utilities/dropdownHelpers.ts
```diff
@@ -0,0 +1,30 @@
...
+
+interface UseDropdown {
+  currentItemIndex: number;
+  dropdownTitle: string;
+  setCurrentItemIndex: React.Dispatch<React.SetStateAction<number>>;
+  setDropdownState: (itemTitle: string, id: number) => void;
+  setDropdownTitle: React.Dispatch<React.SetStateAction<string>>;
+}
+
+export const useDropdown = (
```
mtleonard120, line 11: I dig it.

ClientApp/src/components/reuseables/AssignEmployeesModal/AssignEmployeesModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+  isShowing,
+  isLoading: isLoadingParent,
+  onCancel: onCancelCallback,
+  onConfirm,
+  unassigned,
+  setIsShowing,
+  managers,
+}: AssignEmployeesModalProps): JSX.Element => {
+  const [assigned, setAssigned] = useState<EmployeeOverview[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
```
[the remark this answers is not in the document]
charkour: Yup! Just waiting for the API to be updated.

ClientApp/src/components/reuseables/ConfirmModal/ConfirmModal.tsx
```diff
@@ -3,7 +3,7 @@ import Button from "react-bootstrap/Button";
 import Modal from "react-bootstrap/Modal";
 
 export interface ConfirmModalProps {
-  body: string | JSX.Element;
+  body: React.ReactNode;
```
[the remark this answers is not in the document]
charkour: I'll keep that in mind. That is helpful. Thanks!

ClientApp/src/components/reuseables/AssignEmployeesModal/AssignEmployeesModal.tsx
```diff
@@ -0,0 +1,237 @@
...
+                      );
+                    }}
+                  >
+                    {manager.fullName}
+                  </Dropdown.Item>
+                );
+              },
+            )}
+          </Dropdown.Menu>
+        </Dropdown>
```
jonathanchaffer, line 233: I'm not sure a dropdown is the best UX here. CQL has 50+ employees, so you'd potentially have to scroll a good amount. I would suggest an autocomplete instead.

ClientApp/src/components/reuseables/ConfirmModal/ConfirmModal.tsx
jonathanchaffer, line 6: Reason for using `React.ReactNode` instead of `JSX.Element` here?

ClientApp/src/components/reuseables/AssignEmployeesModal/AssignEmployeesModal.tsx
charkour: Yeah, I'm just following designs.

ClientApp/src/components/reuseables/ConfirmModal/ConfirmModal.tsx
charkour: Yup. I did more research, so if you are expecting children, then it has to be a React.ReactNode because JSX.Element is too strict. In this case, `body` is effectively children. See this [SO Post](https://stackoverflow.com/questions/53688899/typescript-and-react-children-type/53689186#53689186) for more info. Just wrap the ReactNode in a fragment and it'll work as intended: arrays, single nodes, text, undefined, null, booleans.

ClientApp/src/components/reuseables/AssignEmployeesModal/AssignEmployeesModal.tsx
charkour: Yeah, I'm just following designs. But I agree. 
Initially, it was just going to be a list of team leads.

charkour: But as the same time, I don't want to change it if I don't have to.

ClientApp/src/components/reuseables/AutoComplete/AutoComplete.tsx
```diff
@@ -17,6 +17,8 @@ interface AutoCompleteProps<Item> extends FormControlProps {
   getItemValue: (item: Item) => string;
   /** The collection of items to search through. */
   items: Item[];
+  /** (Optional) Callback for when clear button is clicked. */
+  onClearSearch?: () => void;
```
jonathanchaffer, line 21: Nice!

PR #381, opened 2020-08-10
CQLHS-285, CQLHS-148, CQLHS-286

jonathanchaffer:
- Admins can now start the annual review process for an employee
- Admins and team leads can select peer reviewers for an employee
- Once an annual review is started, its information is displayed in the Annual Reviews widget


ClientApp/src/components/pages/EmployeeProfilePage/widgets/AnnualReviewWidget/AnnualReviewWidget.tsx
```diff
@@ -0,0 +1,298 @@
...
+    <Row>
+      <Col>
+        <h4 className="d-inline">{title}</h4>
+      </Col>
+      <Col xs="auto">
+        {openToggle === eventKey ? <ArrowDropUpIcon /> : <ArrowDropDownIcon />}
+      </Col>
+    </Row>
+  );
+};
```
charkour, line 239: Nice. I like how well this works.

ClientApp/src/components/pages/EmployeeProfilePage/widgets/AnnualReviewWidget/StartAnnualReviewModal.tsx
```diff
@@ -0,0 +1,202 @@
...
+  field,
+}: StartAnnualReviewFormSectionProps): JSX.Element => {
+  return (
+    <Row className="mb-3">
+      <Col className="d-flex flex-column">
+        <Form.Label className="d-flex flex-column">
+          <small className="mb-0">{label}</small>
+          <span className="tiny">
+            <em>{hint}</em>
+          </span>
```
charkour, line 178: I think tiny should be bigger. It is incredibly small. I don't think you want to have text smaller than 12px (which is what Google does) on their apps. 8px is too small.

ClientApp/src/components/pages/EmployeeProfilePage/widgets/AnnualReviewWidget/StartAnnualReviewModal.tsx
```diff
@@ -0,0 +1,202 @@
...
+    },
+    [employee, handleError, onHide, onSuccess],
+  );
+
+  const schema = object({
+    finalDeadline: date().required("Final deadline is required."),
+    managerDeadline: date().required("Manager deadline is required."),
+    peerReviewDeadline: date().required("Peer review deadline is required."),
+    selfReviewDeadline: date().required("Self review deadline is required."),
+  });
```
charkour, line 63: I think it would be necessary to validate self <= manager <= peer <= final deadlines

ClientApp/src/components/pages/EmployeeProfilePage/widgets/AnnualReviewWidget/StartAnnualReviewModal.tsx
```diff
@@ -0,0 +1,202 @@
...
+        <Modal.Body>
+          <StartAnnualReviewFormSection
+            error={errors.finalDeadline}
+            field="finalDeadline"
+            hint="Approximate work anniversary"
+            label="Final Deadline"
+            setFieldValue={setFieldValue}
+            touched={touched.finalDeadline}
+            value={values.finalDeadline}
+          />
```
charkour, line 103: in the future, if you wanted you could make an array like this: `const array: {field: string; hint: string; label: string}[]` and map over that creating a section for each one. However, it might not be much fewer lines of code but you can get away with using the field attribute in many places.

ClientApp/src/components/pages/EmployeeProfilePage/widgets/AnnualReviewWidget/AnnualReviewWidget.tsx
```diff
@@ -0,0 +1,298 @@
...
+            <h6>Deadline</h6>
+          </Col>
+          <Col>
+            <h6>Complete?</h6>
+          </Col>
+          <Col>
+            <h6>Action</h6>
+          </Col>
+        </Row>
+      </Card.Header>
```
charkour, line 268: Might consider not showing this if there are no performance reviews.

ClientApp/src/utilities/markingPeriodHelpers.ts
```diff
@@ -1,7 +1,24 @@
...
+    (a: MarkingPeriodSnippet, b: MarkingPeriodSnippet): number => {
+      return moment(b.startDate).valueOf() - moment(a.startDate).valueOf();
+    },
+  );
+};
+
+export const getLastMarkingPeriod = (
+  periods: MarkingPeriodSnippet[] | undefined,
+): MarkingPeriodSnippet | undefined => {
+  return (sortMarkingPeriods(periods) || [])[1];
```
charkour, line 23: How does the returning second marking period equate to the last marking period? Wouldn't it be `const [lastItem] = arr.slice(-1); return lastItem`

ClientApp/src/utilities/markingPeriodHelpers.ts
```diff
@@ -1,7 +1,24 @@
...
+
+export const sortMarkingPeriods = (
+  periods: MarkingPeriodSnippet[] | undefined,
+): MarkingPeriodSnippet[] | undefined => {
+  return periods?.sort(
+    (a: MarkingPeriodSnippet, b: MarkingPeriodSnippet): number => {
+      return moment(b.startDate).valueOf() - moment(a.startDate).valueOf();
+    },
+  );
+};
```
charkour, line 18: Is this sorted by the length of the marking period? Could use some documentation as to what it is sorting by.

ClientApp/src/services/annualReviewsService.ts
```diff
@@ -0,0 +1,124 @@
...
+      return (
+        oldPeerReviewers?.filter((reviewer: AnnualReviwerOverview): boolean => {
+          return reviewer.reviewer?.employeeId === emp.employeeId;
+        }).length === 0
+      );
+    })
+    .map(
+      (emp: EmployeeSnippet): AnnualReviwerOverview => {
+        return {
+          deadline: moment().toISOString(),
```
charkour, line 83: The deadline is today?

charkour: Or is this set later?

charkour:
Also, it would be good to consider keeping subtasks separate in branches since there is a lot going on here.

jonathanchaffer:
> Also, it would be good to consider keeping subtasks separate in branches since there is a lot going on here.

Yup I agree, sorry about that. This branch up like this since the back end was still being updated, and also I couldn't really test some subtasks until others were implemented.

ClientApp/src/components/pages/EmployeeProfilePage/widgets/AnnualReviewWidget/AnnualReviewWidget.tsx
jonathanchaffer: Good call, although any employees with no performance reviews are invalid data, since performance reviews are made when the annual review process starts.

ClientApp/src/services/annualReviewsService.ts
jonathanchaffer: It should be the peer review deadline but there's currently no way to get that. Fix is currently in code review on the back end. Added a TODO

ClientApp/src/utilities/markingPeriodHelpers.ts
jonathanchaffer: Ah, `getLastMarkingPeriod` isn't a good name here. Should be `getPreviousMarkingPeriod`. It's so you can display the details of the marking period that came before the current period.

ClientApp/src/services/annualReviewsService.ts
jonathanchaffer: Update: resolved

jonathanchaffer:
Those changes should be all set, although I'm now getting a 500 on the employees page so this might not be able to be tested yet.

ClientApp/src/utilities/markingPeriodHelpers.ts
charkour: Ah alright, I understand now. Thanks!

PR #385, opened 2020-08-11
CQLHS-454: layout admin dashboard page

charkour:
Widgets are responsive to page size. Columns can vary in height without messing up the other.


charkour:
Build failed; inspecting the issue.

PR #393, opened 2020-08-11
Feature/cqlhs 479

ekm6:
[subtask link](https://cqlcorp.atlassian.net/browse/CQLHS-479)


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/BenchmarkContext.cs
```diff
@@ -166,6 +166,10 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             {
                 entity.Property(e => e.EmployeeId).HasColumnName("EmployeeID");
 
+                entity.Property(e => e.ActiveHireDate)
+                    .HasColumnType("datetime")
+                    .HasDefaultValueSql("('2020-08-11T16:31:11')");
```
tonysulfaro, line 171: I don't think this is expected behavior. Looks you set a default value in the migration and it is going to set the `ActiveHireDate` to when you built the migration project

tonysulfaro: When you create a migration it translates it to SQL and it makes sense that it did that but there might be a diff way to set default datetime values I can look into that

Benchmark.WebApi/Benchmark.WebApi/Benchmark.DbMigrations/Migrations/2020/08/20200811122941_addHireDateToEmployee.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Benchmark.DbMigrations.Migrations._2020._08
+{
+    [Migration(20200811122941)]
+    public class _20200811122941_addHireDateToEmployee : AutoReversingMigration
+    {
+        public override void Up()
+        {
+            Alter.Table("Employee")
+                .AddColumn("ActiveHireDate").AsDateTime().WithDefaultValue(DateTime.UtcNow);
```
tonysulfaro, line 12: Here is what I was referencing earlier, looks like that column will always have a default value of `'2020-08-11T16:31:11'` instead of the current UTC time

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/BenchmarkContext.cs
ekm6: good catch

PR #394, opened 2020-08-11
CQLHS-471 | initial changes to employee controller, add validation

tonysulfaro:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Request/EmployeeUpdateRequest.cs
```diff
@@ -1,16 +1,24 @@
...
+        [Required, Id]
         public int EmployeeId { get; set; }
+        [Required]
         public string Email { get; set; }
+        [Required]
         public string FirstName { get; set; }
+        [Required]
         public string LastName { get; set; }
         public bool? IsActive { get; set; }
         public int? ManagerId { get; set; }
```
ekm6, line 18: could you double check the logic for when this is null?

tonysulfaro: yeah it blows up so making it required solves that

PR #406, opened 2020-08-12
CQLHS-496 | initial work on file deletion + frontend validation fixes

tonysulfaro:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Database/Milestone.cs
```diff
@@ -10,7 +10,7 @@ public Milestone()
         {
         }
 
-        [Required, Id]
+        [Required]
```
ekm6, line 13: needs ID

PR #408, opened 2020-08-12
CQLHS-487: fix elements missing keys

jonathanchaffer:
In commit e10de9e, I regenerated types and reverted any changes in `api.ts` that weren't directly related to this, since they caused compiler errors which are being fixed on a different ticket.

There are some errors for duplicate keys, this is due to invalid data not missing keys on the front end.


jonathanchaffer:
Problem is the objects are _exact_ duplicates, so `JSON.stringify` would still produce duplicate keys. I think the duplication problems are already fixed on the back end, there's just some leftover duplicates from when it used to be broken.

charkour:
okay, that's annoying. Thanks for the explanation.

PR #424, opened 2020-08-13
CQLHS-506 | bug fix to address project created by and date being null

tonysulfaro:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Request/ProjectUpdateRequest.cs
```diff
@@ -24,6 +24,12 @@ public class ProjectUpdateRequest
 
         public DateTime? EvaluationDeadline { get; set; }
 
+        [Required, Id]
```
ekm6, line 27: needs to be nullable

PR #432, opened 2020-08-14
CQLHS-516 | remove required from position

tonysulfaro:


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Models/Request/PositionUpdateRequest.cs
```diff
@@ -28,7 +28,6 @@ public PositionUpdateRequest(EmployeeRole er)
         [Required, Id]
```
jonathanchaffer, line 28:
```suggestion
```
`EmployeeRoleId` should also be optional

PR #449, opened 2020-08-19
CQLHS-491 | modify email template for annual review reminder

tonysulfaro:
https://cqlcorp.atlassian.net/browse/CQLHS-491


Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmployeeService.cs
```diff
@@ -178,12 +178,12 @@ public async Task<OperationResult> Remind(AnnualReviewReminderRequest request)
...
+                //emailservice.Send(request.Email, "Annual Review Reminder", emailservice.AnnualReviewReminder(request.EmployeeUnderReviewName, request.Deadline.ToString()));
+                emailservice.Send("enoch.mwesigwa@cqlcorp.com", "Annual Review Reminder", emailservice.AnnualReviewReminder(request.EmployeeUnderReviewName, request.Deadline.ToString()));
+                emailservice.Send("tony.sulfaro@cqlcorp.com", "Annual Review Reminder", emailservice.AnnualReviewReminder(request.EmployeeUnderReviewName, request.Deadline.ToString()));
+                emailservice.Send("Charlie.Kornoelje@cqlcorp.com", "Annual Review Reminder", emailservice.AnnualReviewReminder(request.EmployeeUnderReviewName, request.Deadline.ToString()));
+                emailservice.Send("jon.chaffer@cqlcorp.com", "Annual Review Reminder", emailservice.AnnualReviewReminder(request.EmployeeUnderReviewName, request.Deadline.ToString()));
+                emailservice.Send("ally.brooks@cqlcorp.com", "Annual Review Reminder", emailservice.AnnualReviewReminder(request.EmployeeUnderReviewName, request.Deadline.ToString()));
```
mtleonard120, line 186: Should lines 182-186 be removed and line 181 be uncommented here?

Benchmark.WebApi/Benchmark.WebApi/Emails/email-no-buttons.html
```diff
@@ -0,0 +1,344 @@
+<!DOCTYPE html>
```
mtleonard120, line 1: Interesting - is this like the non-stringified version of the email html? How is this being used?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmailService.cs
```diff
@@ -88,16 +88,17 @@ public string ProjectEvaluationModified(string projectName, string primaryLink)
...
...ing=\"0\" style=\" width: 100%; padding-top: 4em; padding-bottom: 2em; padding-left: 0; padding-right: 0; \" > <tbody> <tr> <td> <table class=\"container\" cellpadding=\"0\" cellspacing=\"0\" style=\" width: 100%; max-width: 500px; margin-left: auto; margin-right: auto; padding-left: 2em; padding-right: 2em; \" > <tbody> <tr> <td> <h1 style=\"margin-top: 0; font-weight: 600;\"> Header </h1> <p class=\"preheader\" style=\"margin-top: 0; margin-bottom: 2em;\" > email body. </p><h4 style=\"margin-bottom: 0.5em; font-weight: 700;\"> Content subhead1: </h4> <p style=\"margin-top: 0; margin-bottom: 2em;\"> Lorem ipsum dolor sit amet1. </p><h4 style=\"margin-bottom: 0.5em; font-weight: 700;\"> Content subhead2: </h4> <p style=\"margin-top: 0; margin-bottom: 2em;\"> Lorem ipsum dolor sit amet2. </p><table class=\"card-footer\" cellpadding=\"0\" cellspacing=\"0\" style=\" width: 100%; padding-top: 0.75em; padding-bottom: 0.75em; \" > <tbody> <tr> <td class=\"spacer\" style=\"width: 100%;\"></td><td class=\"right\" style=\" margin-right: 0; white-space: nowrap; \" ></td></tr></tbody> </table> </td></tr></tbody> </table> </td></tr></tbody> </table> </td></tr></tbody> </table> </body></html>";
```
mtleonard120, line 93: Looks like a ton got cut from this string - just curious, what happened here?

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmployeeService.cs
tonysulfaro: I think this is more of a go-live scenario when development is subsiding and users are ready to start using the application. If it were uncommented then it would send out emails to real employees. We can discuss what the plan is for deployment is!

mtleonard120: Ah I see - would it be possible to check the request for whether it's coming from the prod site or not? In prod instances you would want the real emails to send but from any other origin you would want just dev emails to send.

Benchmark.WebApi/Benchmark.WebApi/Emails/email-no-buttons.html
tonysulfaro: Currently, the emails stored in the `Benchmark.WebApi` folder are just for reference but the build server scripts will eventually be updated to copy them over to where the app is running so it can read them in through file io instead of being hardcoded like they are now.

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmployeeService.cs
tonysulfaro: that sounds like a good idea, I'll look into it but I'm sure there is some best practice that I am not aware of atm

Benchmark.WebApi/Benchmark.WebApi/Benchmark.WebApi/Services/EmailService.cs
tonysulfaro: This was actually using the bootstrap email template and got switched over to the one Jon created so its actually much smaller than the generated one

